PR #19, opened 2024-06-20
fix: _regenerateKeys so that the MarkDefs left are only those allowed

RitaDias:
### Description

Add fix for when PTE text is pasted from one input to another that it respects the allowed annotations

Before

https://github.com/sanity-io/sanity/assets/6951139/a647261b-aab8-4476-ac13-8e647528d5fe

### Testing

The added tests should be sufficient
- Added tests to `_regenerateKeys` (test before normalization)
- The normalization tests already test for the "after normalization" -> `preserves other marks that apply to the spans`

While the original issue was detected on copy paste, I don't think it is necessary to add a test to target that use case in particular. We would have seen this as an issue when creating a doc via the terminal for example.

### Notes for release

Fixes issue where when PTE text was copy pasted always kept the original annotations regardless of the receiving PTE customization


packages/editor/src/editor/plugins/createWithInsertData.ts
```diff
@@ -273,11 +274,35 @@ function _regenerateKeys(
...
+        return !annotations.includes(def._type)
+      })
+
+      // if they have forbidden annotations, we remove them and keep the rest
+      if (hasForbiddenAnnotations) {
+        const allowedAnnotations = (newNode.markDefs || []).filter((def) => {
+          return annotations.includes(def._type)
+        })
+
+        return {...newNode, markDefs: allowedAnnotations, _key: keyGenerator()}
```
christianhg, line 303: I find this function a bit confusing to read since this if clause now either does an early return of manipulates `newNode` directly. But maybe that's not something we should refactor unless the rest of the function's branches are covered by tests as well.

christianhg (approved):
Nice! I really like the test cases added.

packages/editor/src/editor/plugins/createWithInsertData.ts
RitaDias: I think it does what it is expected,
the new node will always be modified directly (even in the old part of the code that wasn't changed). The point is to make sure that whatever node is returned from this has the correct MarkDefs.

What I am wondering now is if in the case where there are two annotations, one is allowed and one isn't if the allowed one should just be allowed to proceed or should receive the same treatment as the line bellow since as the code is now that will not happen now that I think about it 🤔 @skogsmaskin thoughts?

christianhg: True, I too think it does what is expected. I just found it a tad hard to read and understand the logic 😇

christianhg: > What I am wondering now is if in the case where there are two annotations, one is allowed and one isn't if the allowed one should just be allowed to proceed

I think all allowed annotations should be preserved. Good point! We probably need a test case for that and adjust the logic. If you are busy then I can give it a try for you if you like :)

RitaDias: > True, I too think it does what is expected. I just found it a tad hard to read and understand the logic 😇

It's an absolutely fair point, should have definitely added a comment about it to make it clearer :)

RitaDias: > > What I am wondering now is if in the case where there are two annotations, one is allowed and one isn't if the allowed one should just be allowed to proceed
> 
> I think all allowed annotations should be preserved. Good point! We probably need a test case for that and adjust the logic. If you are busy then I can give it a try for you if you like :)

I'm currently in the middle of something else, if you have the time I'd be super thankful since i'll likely only have time on Monday 🙏

christianhg: I'll give it a try! I really should get more familiar with this codebase anyway.

packages/editor/src/editor/plugins/__tests__/createWithInsertData.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+    expect(generatedValue).toStrictEqual([
+      {
+        _key: '',
+        _type: 'myTestBlockType',
+        children: [
+          {_key: '', _type: 'span', marks: [''], text: 'Block A'},
+          {
+            _key: '',
+            _type: 'span',
+            marks: [''],
```
christianhg, line 72: Just so I understand, is this actually what we expect? Why is `marks: ['']` in this test case? Shouldn't we preserve all marks in this teset?

packages/editor/src/editor/plugins/createWithInsertData.ts
RitaDias: If you need help let me know I can try my best even if I myself don't have a lot of context :)

packages/editor/src/editor/plugins/__tests__/createWithInsertData.test.tsx
RitaDias: Ah this is because of my mock for the generateKey 
     ` jest.fn().mockReturnValue('') as () => string,`

Because I wasn't concerned about what key it was, I just wanted to make sure it was spitting the right number of marks and markDefs, the value of the keys themselves don't really matter

That being said, probably best to do something like 

     ` jest.fn().mockReturnValue('dummy') as () => string,`

on line 57

RitaDias: Mentioned it in the comment at line 62 :)

PR #34, opened 2024-06-26
fix: reset selection on blur

christianhg:
https://github.com/portabletext/editor/assets/4602382/94caf8aa-5d86-4398-9d14-910a402eb543

Make a selection:
Screenshot 2024-06-26 at 15 34 46

And then blur the editor:
Screenshot 2024-06-26 at 15 34 19

Before this change, the selection would be persisted after blurring the editor. This would make the toolbar still function:
Screenshot 2024-06-26 at 15 34 48

The problem is (unsurprisingly) also present in the Studio (notice the selected "B" in the blurred editor):
Screenshot 2024-06-26 at 15 37 23

### DIFF
--- a/packages/editor/src/editor/components/Synchronizer.tsx
+++ b/packages/editor/src/editor/components/Synchronizer.tsx
@@ -124,6 +124,16 @@ export function Synchronizer(props: SynchronizerProps) {
           })
           onChange(next) // Keep this out of the startTransition!
           break
+        case 'blur':
+          // Set the selection state in a transition, we don't need the state immediately.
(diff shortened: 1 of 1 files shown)


skogsmaskin (commented):
There could be an issue here I think in regard to embedded dialogs and similar.

For instance, if you open a modal from a void block, you would want the void block to still be selected in the editor. So when you close the modal, you return to where you were (void block selected) without having to keep track of the previous selection. I have a feeling that this will completely wipe out that selection and put the selection to the top of the editor when it's refocused?

PR #36, opened 2024-06-26
fix(deps): upgrade debug module to align with sanity repo

rexxars:
The sanity project uses the debug module at version `^4.3.4`, while the editor is using `^3.2.7`. There aren't any breaking changes between these two apart from very old node versions not being supported anymore. By aligning on the version range, we reduce the amount of duplicate dependencies in our dependency tree.


christianhg (commented):
Thank you!

PR #37, opened 2024-06-26
test: upgrade test-utils to latest version

rexxars:
As part of filing #36, I ran the tests and saw there were some warnings about using the deprecated `act` from the testing library. To silence this warning, I modified the code to import from react, but because the testing library is also _internally_ calling `act` from `react-dom`, it still gave warnings.

I looked through the `@testing-library/react` change logs for the different major versions since the one we're currently on (13), and didn't see anything that looked too dangerous (see [v14](https://github.com/testing-library/react-testing-library/releases/tag/v14.0.0) + [dom-testing-library@v9](https://github.com/testing-library/dom-testing-library/releases/tag/v9.0.0), [v15](https://github.com/testing-library/react-testing-library/releases/tag/v15.0.0) and [v16](https://github.com/testing-library/react-testing-library/releases/tag/v16.0.0)).

Interestingly, upgrading does silence the warnings, but yields errors in the tests. I don't know if these are actually real bugs or an issue with the tests, but I thought it would be worth raising a PR to highlight this and perhaps have you investigate!

For what its worth, it seems to be the upgrade to v14 (+ dom-testing-library@v9) that breaks the tests.


christianhg:
Thank you for this. I did go down the exact same path to try and silence this warning, and when I just ended up with all tests failing I decided to pause my efforts. But it's good to have an open PR for this so we can get it fixed.

christianhg:
@SocketSecurity ignore-all

PR #65, opened 2024-07-04
fix(editor): allow returning Promise(undefined) in paste handler

christianhg:
Before this change, nothing would happen if you returned `undefined` wrapped in a `Promise` from the paste handler. Now, we treat this as a signal to fall back and paste normally by passing `event.clipboardData` to the internal `.insertData(...)` method. This resembles how the paste handler control flow functioned before https://github.com/portabletext/editor/pull/28. However, returning `Promise(undefined)` is not encouraged since it doesn't do what you expect in Firefox. In Firefox, the `clipboardData` isn't available anymore after the Promise microtask.

### DIFF
--- a/packages/editor/src/editor/Editable.tsx
+++ b/packages/editor/src/editor/Editable.tsx
@@ -404,7 +404,10 @@ export const PortableTextEditable = forwardRef(function PortableTextEditable(
         Promise.resolve(onPasteResult)
           .then((result) => {
             debug('Custom paste function from client resolved', result)
-            if (result && result.insert) {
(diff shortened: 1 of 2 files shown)


christianhg:
Her is an example of a paste handler that doesn't work anymore after #28 : https://github.com/sanity-io/www-sanity-io/blob/main/packages/sanity-config/src/components/blockEditor/handlePaste.ts#L26

It should be noted that this paste handler could easily be turned into a synchronous one, and then it would work, but I don't know how many paste handlers returning `Promise(undefined)` exist in the wild.

PR #81, opened 2024-07-30
Adjust package config

christianhg:


packages/editor/package.config.ts
```diff
@@ -20,10 +20,7 @@ export default defineConfig({
     ],
     rules: {
       // Disable rules for now
-      'ae-forgotten-export': 'off',
       'ae-incompatible-release-tags': 'off',
```
christianhg, line 23: @stipsan , I thought about removing this one as well because it would just be nice if the package functioned without these suppressions. However, it seems that this suppression is needed because the `PortableTextEditableProps` is marked `@public` even though most (all?) of its props are marked `@beta` or `@alpha`.

What would be the correct thing to do here?

Screenshot 2024-07-30 at 11 00 28

stipsan: `ae-forgotten-export` should at least be `warn`.
While `ae-incompatible-release-tags` is so often too pedantic for us and this wouldn't be the first place we have to disable it 😅

stipsan (approved):

packages/editor/package.config.ts
christianhg: Thanks!

PR #86, opened 2024-08-01
chore: add react compiler playground

stipsan:
This PR adds a new `react-compiler` playground and studio embed.
It takes the [existing playground](https://portable-text-playground-git-add-react-compiler-playground.sanity.dev/), and [reuse it in a Next v15 environment](https://editor-react-compiler-git-add-react-compiler-playground.sanity.dev/playground), which runs the code on React 19 RC. It also runs the React Compiler on the codebase of `@portabletext/editor`, allowing us to quickly test if doing so breaks it in any way, or if it's fine when it's optimized this way.

It also [takes the Sanity Studio test setup](https://editor-sanity-studio-git-add-react-compiler-playground.sanity.dev), and [embeds it as well](https://editor-react-compiler-git-add-react-compiler-playground.sanity.dev/studio).

As the `@portabletext/editor` already works quite well on React 19 by itself I didn't bother adding tests for that, while there's work to be done for it to be ready for React Compiler, as well as React 18's Concurrent rendering features (the Studio team is very interested in starting to use `useDeferredValue` and for that to be viable we have work to do to ensure renders are pure and we remove the cases where `useMemo` are used to perform side-effects that would be better done in a `useLayoutEffect` 🙌 ).


apps/sanity-studio/package.json
```diff
@@ -28,8 +28,8 @@
   },
   "devDependencies": {
     "@sanity/eslint-config-studio": "^4.0.0",
-    "@types/react": "^18.0.25",
+    "@types/react": "^18.3.3",
     "eslint": "^8.6.0",
-    "typescript": "^5.1.6"
+    "typescript": "5.5.4"
```
christianhg, line 33: Why pinning?

stipsan: TypeScript doesn't follow semantic versioning and there are often breaking changes on minor version bumps: https://github.com/microsoft/TypeScript/issues/14116

christianhg: Is it a problem? The renovate bot PR will just fail in that case and then we can fix the type errors in that PR before it's merged.

apps/react-compiler/turbo.json
```diff
@@ -0,0 +1,16 @@
...
+  "$schema": "https://turbo.build/schema.json",
+  "extends": ["//"],
+  "tasks": {
+    "build": {
+      "env": ["DEBUG"],
+      "outputs": [".next/**", "!.next/cache/**"],
+      "dependsOn": ["^build"]
+    },
+    "dev": {
+      "dependsOn": ["^build"],
```
christianhg, line 11: This is a good config that we probably want to move to the top-level `turbo.json`. And then this local `dev` task can be removed.

christianhg (approved):
Awesome 👏

apps/sanity-studio/package.json
stipsan: Yes, we've had problems in the past where `@sanity/pkg-utils` would type check differently than `tsc --noEmit` due to different TS versions sneaking into the codebase that are all within the same major, but on different minors. Being explicit ensures that if anyone in the future were to nuke the `pnpm-lock.yaml` file and regen it, then you won't suddenly have build errors that you didn't have before, just because a new minor of TypeScript is pulled in that haven't been tested yet.
The philosophy is that doing things like nuking the lockfile at any time on `main` should be safe to do and not introduce unexpected behavior. And AFAIK that requires `typescript` to be pinned 😅

christianhg: That makes sense. But does that then mean that we need to bump `typescript` manually from now on?

stipsan: Renovate handles it ☺️

christianhg: OK, I got no problem then 😆

PR #214, opened 2024-09-25
feat(editor): add store prop 

christianhg:
The `store` is a new, experimental way of interacting with and listening to
events from the editor.

So far it's only used to send patches into the editor.


packages/editor/src/editor/PortableTextEditor.tsx
```diff
@@ -105,26 +107,46 @@ export class PortableTextEditor extends Component<PortableTextEditorProps> {
...
     this.change$.next({type: 'loading', isLoading: true})
 
     this.schemaTypes = getPortableTextMemberSchemaTypes(
       props.schemaType.hasOwnProperty('jsonType')
         ? props.schemaType
         : compileType(props.schemaType),
     )
+
+    // If no store is provided then we just create one
+    this.store = props.store ?? createEditorStore()
```
EoinFalconer, line 131: The word store here is the only thing I can think could be discussed. I prefer the API – I just don't know about the terminology.

christianhg: Thanks! I've renamed it to `editor` now.

PR #319, opened 2024-10-30
feat: use react-compiler to optimise render

stipsan:
[Test PR in the Studio](https://github.com/sanity-io/sanity/pull/7706)


packages/editor/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@portabletext/editor",
-  "version": "1.1.11",
+  "version": "1.1.12-canary.1",
```
christianhg, line 3: Why canary? What complications will this bring to the ordinary release process?

packages/editor/package.json
```diff
@@ -91,10 +93,16 @@
...
     "@types/react-dom": "^18.3.1",
     "@types/ws": "~8.5.12",
+    "@typescript-eslint/eslint-plugin": "^8.12.2",
+    "@typescript-eslint/parser": "^8.12.2",
     "@vitejs/plugin-react": "^4.3.3",
     "@vitest/browser": "^2.1.4",
     "@xstate/react": "^4.1.3",
+    "babel-plugin-react-compiler": "beta",
     "dotenv": "^16.4.5",
+    "eslint": "8",
```
christianhg, line 103: Guess who's back 😂

packages/editor/package.json
stipsan: No complications, release please ignores the version used here and follows the one in the manifest.
I kept it in the version field for transparency on the fact that I published a canary of this PR 😌

packages/editor/package.json
stipsan: Well, it doesn't seem like biome does a good job validating useEffect deps? 😅

christianhg:
Is it perhaps a bit incorrect to use `feat` for this commit? As far as I can tell, the PR enables the compiler for the Playground and the tests. Or did I misunderstand something?:

packages/editor/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@portabletext/editor",
-  "version": "1.2.0",
+  "version": "1.1.12-canary.3",
```
christianhg, line 3: Do we want to merge with this change? I don't understand...

christianhg (changes requested):
I left some doubts in the comments so marking this as "Request changes"

christianhg (approved):

packages/editor/package.json
stipsan: Release please will correct it later so it's fine :)

christianhg:
LFG

PR #338, opened 2024-11-01
fix(react-compiler): handle mutation violation

stipsan:


christianhg:
I'm unsure about this change and can't really foresee the consequences. I would need to test range decorations to make sure everything still works as before. Or did you already do that?

stipsan:
> I'm unsure about this change and can't really foresee the consequences. I would need to test range decorations to make sure everything still works as before. Or did you already do that?

I did, yep!

christianhg (approved):

PR #353, opened 2024-11-03
fix: move `@xstate/react` from dev to production dependency

rexxars:
Note: This is currently breaking `sanity@next`.

[This line](https://github.com/portabletext/editor/blob/995bbe5e43f62061c4ebe8a71403c26c2a3acbea/packages/editor/src/editor/use-editor.ts#L6) is importing from `@xstate/react`, which is marked as a dev dependency - not a production dependency, so it is not installed for most applications.

### DIFF
--- a/packages/editor/package.json
+++ b/packages/editor/package.json
@@ -57,6 +57,7 @@
   },
   "dependencies": {
     "@portabletext/patches": "1.1.0",
+    "@xstate/react": "^4.1.3",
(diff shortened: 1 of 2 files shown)


christianhg (approved):
Argh! Sorry about that! Originally I didn't want this package as a dependency at all, so that might be the reason this mistake crept in. `useEditor` is still experimental.

PR #374, opened 2024-11-05
fix: remove @sanity/util dependency

christianhg:
By copy/pasting the source code of `randomKey`.


stipsan (approved):
Makes sense!

christianhg:
> Makes sense!

Thanks!

PR #414, opened 2024-11-13
fix: refactor `PortableTextEditor` from class component to functional component

stipsan:
This allows it to be optimized, and helps us better guard against race conditions and lifecycle issues.

### DIFF
--- a/apps/playground/src/portable-text-toolbar.tsx
+++ b/apps/playground/src/portable-text-toolbar.tsx
@@ -4,6 +4,7 @@ import {
   usePortableTextEditorSelection,
   type Editor,
   type EditorSelection,
+  type PortableTextEditorInstance,
 } from '@portabletext/editor'
 import {SquareDashedMousePointerIcon} from 'lucide-react'
 import {isValidElement, useMemo} from 'react'
@@ -105,7 +106,7 @@ export function PortableTextToolbar(props: {
 
 function StyleSelector(props: {
   schemaDefinition: SchemaDefinition
-  editor: PortableTextEditor
+  editor: PortableTextEditorInstance
   selection: EditorSelection
 }) {
   const focusBlock = PortableTextEditor.focusBlock(props.editor)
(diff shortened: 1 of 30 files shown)


packages/editor/src/editor/PortableTextEditor.tsx
```diff
@@ -1,460 +1,314 @@
...
+    useImperativeHandle(forwardedRef, () => ref.current!)
 
-  // Static API methods
-  static activeAnnotations = (
-    editor: PortableTextEditor,
+    /**
+     * Unclear if `editorRef` is really needed, as setting a regular `ref` has the same effect (and always has been)
+     */
+    useImperativeHandle(props.editorRef, () => ref.current!)
+
```
stipsan, line 127: I don't think we need `editorRef` @christianhg, as we can achieve the same thing with just `ref` 🙌

apps/playground/src/portable-text-toolbar.tsx
```diff
@@ -105,7 +106,7 @@ export function PortableTextToolbar(props: {
 
 function StyleSelector(props: {
   schemaDefinition: SchemaDefinition
-  editor: PortableTextEditor
+  editor: PortableTextEditorInstance
```
christianhg, line 109: This PR looks promising, but I think it's a breaking change (both in PTE but also for Studio users) that you can't reference the `PortableTextEditor` as a type anymore.

In any case, this PR gave me an idea for a different take on how to tackle this problem. I'll try to open another PR to compare with.

stipsan: It's always been a nasty anti pattern given that it's a react class component. If you can get it working by reducing the class component to a very thin layer without logic (and preferably use PureComponent instead of Component since the React Compiler only supports function components and will never add class components) then that could work I suppose 🤔

christianhg:
Thank you for this PR. I have a similar idea I'm pursuing (with the same end goal, but without breaking changes), so I'll close this PR for now and hopefully open one similar soon.

PR #543, opened 2024-12-09
docs(behavior): initial draft

christianhg:


packages/editor/README.md
```diff
@@ -370,14 +370,328 @@ function Toolbar() {
...
+
+```ts
+import type {EditorSelector} from '@portabletext/editor/selectors'
+
+const selectionIsBackwards: EditorSelector<boolean> = ({context}) => {
+  // ...
+}
+```
+
+This behavior also shows another neat trick:
```
robinpyon, line 537: nit: I think calling it a 'trick' is a little reductive! Perhaps it could be omitted entirely

packages/editor/README.md
```diff
@@ -370,14 +370,328 @@ function Toolbar() {
...
+
+> 💡 You can return **parameters** from the `guard` that you can reuse when firing actions
+
+This is extremely useful since you often want to perform a set of actions usings state selected in the guard. Instead of having to pluck that state again in the action callback, you can just return and reuse it.
```
robinpyon, line 541:
```suggestion
This is extremely useful since you often want to perform a set of actions using state selected in the guard. Instead of having to pluck that state again in the action callback, you can just return and reuse it.
```

packages/editor/README.md
```diff
@@ -370,14 +370,328 @@ function Toolbar() {
...
+```
+
+`selectionIsCollapsed` and `getFocusListBlock` are just pure functions (called "selectors") that derive state from this context. Both are built-in and exported from `@portabletext/editor/selectors` (along with a number of other useful, built-in selectors).
+
+```ts
+import {
+  getFocusListBlock,
+  selectionIsCollapsed,
+} from '@portabletext/editor/selectors'
+```
```
robinpyon, line 525: nit: when referencing the above example, my first question was "What is selectionIsCollapsed – is it exported, or just a throwaway helper for this example?" 

I wonder if it'd make sense to include the import declarations in the above example (which immediately signals these are 'official') whilst keeping the blurb here on what these selectors are (which is a fantastic segue into BYO selectors)

packages/editor/README.md
```diff
@@ -370,14 +370,328 @@ function Toolbar() {
...
+
+```tsx
+<button
+  onClick={() => {
+    props.editor.send({
+      type: 'decorator.toggle',
+      decorator: 'strong',
+    })
+    props.editor.send({type: 'focus'})
```
robinpyon, line 665: mega nit: perhaps we don't need to include the `props.` here, since whether that editor instance is being passed to this component or not is irrelevant to this example

packages/editor/README.md
```diff
@@ -370,14 +370,328 @@ function Toolbar() {
...
+- `delete.text`
+- `effect`
+- `reselect`
+- `select`
+- `select.previous block`
+- `select.next block`
+- `style.add`
+- `style.remove`
+- `text block.set`
+- `text block.unset`
```
robinpyon, line 694: nit: alphabetically sort?

robinpyon (commented):
As a layperson the section on the behavior API was very easy to understand. I've outlined a few nits in comments but otherwise I think it's great!

---

I do have some other general feedback on this README (beyond the Behavior API additions) which are listed below. Feel free to ignore!

- The README probably doesn't need to lead with the section on **End-User experience**. Whilst we should celebrate the testing story, it's probably not the first thing people are after here.
- Beyond the initial disclaimer (which is really thorough and considerate!), I would expect the first section to be a 'Quickstart' example. This would be the smallest possible example we can provide. Even if we don't yet have a story of how users can consume this outside of Sanity, a pithy example can communicate the general shape of the API incredibly quickly 
- In the section on **Render Marks, Blocks and Objects**, the callout on lists ("However, lists are a little special...") is welcome, but admittedly a little distracting. Maybe it would be good to expose the caveat on lists via a collapsible `<details>` element and keep the emphasis on the happy path
- Additionally, the three examples listed here (**Render the Editor Component**, **Render Marks, Blocks and Objects** and **Render the Toolbar**) run a little long! I think there's potential here to simplify these, or even increase the line length so it's not so laborious to skim through. For example, the **Render Marks...** example may not need to be so exhaustive (maybe it just handles two heading styles instead of four, or one annotation, etc)

- In the section on 'Render the Toolbar', it may be worth calling out that rendering a toolbar is completely optional!

PR #586, opened 2024-12-18
docs: update readme with new docs content and links

markmichon:
Updates readme to match new content.
- Pulls in the getting started guide's content to the readme
- Adds links out to the official docs for each section.
- Summarizes behavior api and directs out to docs.


packages/editor/README.md
```diff
@@ -180,216 +175,185 @@ const renderStyle: RenderStyleFunction = (props) => {
...
+  }
+  if (props.value === 'underline') {
+    return <u>{props.children}</u>
+  }
+  return <>{props.children}</>
 }
 ```
 
-### Render the Toolbar
+_Note: By default, text is rendered as an inline `span` element in the editor. While most render functions return a fragment (`<>`) as the fallback, make sure block level elements return blocks, like `<div>` elements._
```
kmelve, line 205: Make this a github markdown note block thingy?

packages/editor/README.md
```diff
@@ -180,216 +175,185 @@ const renderStyle: RenderStyleFunction = (props) => {
...
-      onClick={() => {
-        editor.send({type: 'style.toggle', style: props.style})
-        editor.send({type: 'focus'})
-      }}
-    >
-      {props.style}
-    </button>
-  )
-}
+### View the Portable text data
```
kmelve, line 335:
```suggestion
### View the Portable Text data
```

kmelve (changes requested):
All over great! I realize we should do a bit more "marketing" at the top where we unpack the value and "features" of PTE, but we can do that later.

kmelve (approved):
🚢

PR #606, opened 2024-12-18
Docs/smol fixes

kmelve:
Not 100% on using the graphic vs the logo (if we use the logo… it should be bigger)


markmichon:
Yeah I don't love the way images sit in the hero theme. Not personally attached to either though.

markmichon (approved):

PR #631, opened 2024-12-19
wip: annotated typedocs progress

markmichon:
Quite a few misc complications still with the typedocs. Not entirely sure if the problem is how we do things, how they assume people do things, or something in between. 

I did discover that `pkg-utils` is blocking our ability to wholesale use all the typedocs tags. Should be able to manually add them via `package.config.ts` until we hit an inline tag. At the point the library will need an update since the type only supports `block` and `modifier` right now.

If the groups or categories will actually work, I think we'll be in good shape, but this implementation is still just "okay".

### DIFF
--- a/apps/docs/astro.config.mjs
+++ b/apps/docs/astro.config.mjs
@@ -3,7 +3,13 @@ import starlight from '@astrojs/starlight'
 import tailwind from '@astrojs/tailwind'
 import {defineConfig} from 'astro/config'
 import starlightLinksValidator from 'starlight-links-validator'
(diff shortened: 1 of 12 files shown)


markmichon:
The [createMarkdownBehaviors](https://portable-text-editor-documentation-mf1kpj5m5.sanity.build/api/behaviors/functions/createmarkdownbehaviors/) page is a good example of the bizarre way it handles some properties.

markmichon:
CleanShot 2025-01-07 at 11 45 08@2x
Looking a bit better.

PR #657, opened 2025-01-03
docs: document `raise` action

christianhg:


markmichon:
@christianhg This is great! I've split the details from this into the reference, concept, and recipe pages. I'll see if there's a nice way to annotate all the actions and events directly with the typedef docs once that's more mature.

markmichon (approved):

christianhg:
> @christianhg This is great! I've split the details from this into the reference, concept, and recipe pages. I'll see if there's a nice way to annotate all the actions and events directly with the typedef docs once that's more mature.

Thanks a lot for the review and the revisions made! 🙇

PR #691, opened 2025-01-13
docs: add more details to typedoc support

markmichon:
Updates the typedoc setup to be a bit cleaner. This should allow us to annotate in code, and have somewhat stable reference links.


christianhg (approved):
Very nice 👏

PR #784, opened 2025-02-09
feat: add support for Before and After hooks

aslakhellesoy:
This adds support for Before and After hooks.

Cucumber allows for scoping of hooks via tags - I have not added this in this PR.

Hooks are useful for setting up context (such as clearing a database).


christianhg (approved):
This is so good! I've wanted to add these, but didn't get around to it yet. Thank you!

I hope it's alright with you that I put the hooks test in a separate file. This way, we've preserved the Data Table test that you created in the previous PR.

aslakhellesoy:
Thanks so much for merging @christianhg! I agree it's better to extract this to a separate tests.

PR #980, opened 2025-04-02
test: add failing test for react key warning on value update

ricokahler:
### Summary

This PR introduces a new test case to the `packages/editor` suite. This test currently **fails** and serves to reproduce and highlight an existing issue where React may log duplicate key warnings (`Encountered two children with the same key...`) during certain `update value` operations in the Portable Text Editor. This scenario is particularly relevant to how our SDK team utilizes the editor.

### Observed Behavior & SDK Context

When the editor's value is updated via `editor.send({type: 'update value', value: ...})`, under specific conditions, React logs a warning about duplicate keys.

This behavior is triggered by a pattern relevant to our SDK implementation:
*   The SDK reuses nodes (Portable Text block/span objects) from the previous state when constructing the new value for an update. This is done for performance reasons and allows for efficient change detection using strict equality (`===`) checks, as we follow immutable update patterns.
*   The issue manifests when a *parent* block node is reused (has the same `_key` and potentially the same object identity), but its *child* content (like the `text` in a span) is modified.
*   In this case, React logs the key warning, potentially indicating an unexpected interaction between the editor's update mechanism (when handling reused parent nodes with changed children) and React's component reconciliation process based on keys.

Here is video demonstrating this issue within the SDK's context. The code running locally is on the left. Notice how when I update the value remotely, it shows the duplicate key error.

This is an issue because it causes the editor to be out of sync and it doesn't seem to recover. It remains out of sync after the edit.

https://github.com/user-attachments/assets/06d34623-2e33-4e3b-8fa9-adba21c2c06c

### Test Case Details

A new test case, `should allow updating block text content without React key warnings`, has been added to `packages/editor/tests/event.update-value.test.tsx`.

This test specifically replicates the SDK's update pattern where the warning occurs:
1.  Initializes the editor with an empty value.
2.  Performs an `update value` with two blocks (`block1`, `block2`).
3.  Waits briefly (using `setTimeout` with 0ms to allow the event loop to process).
4.  Performs a second `update value` where the `block1` object is intentionally constructed to simulate reuse (same `_key`), but its `children` array contains a span with modified `text`. The `block2` object remains the same. This mirrors the SDK modifying nested content within a reused block structure.
5.  Spies on `console.error` using `vi.spyOn`.
6.  Asserts that the specific React key warning (`Encountered two children with the same key...`) was *not* called, reflecting the desired behavior.

### Verification

This test currently **fails** as expected. The assertion fails because the underlying issue *does* cause the React key warning to be logged to the console during the second `update value` call, confirming the test successfully reproduces the problematic behavior encountered by the SDK. This provides a concrete case for investigating the root cause.


christianhg:
Thank you for reporting this, @ricokahler , and thank you for the very helpful failing test case. The problem had to do with the internal sync logic potentially deleting empty spans if the same span suddenly had text. I've added a check before the deletion and also updated your test case to reflect the desired outcome.

PR #983, opened 2025-04-03
feat: introduce `fields` on schema types defined using `defineSchema`

christianhg:
There are two ways to configure the schema used in PTE:

1. `<EditorProvider schema={someSanitySchema}>`
2. `<EditorProvider schemaDefinition={defineSchema(...)}>`

The first option is the old way of configuring the schema and is used in Studio, Create and Canvas.

The second option was added to provide a user-friendly and Sanity-agnostic way to configure a simple schema without having to deal with Sanity opinions on default fields added to object types with certain names and without having to deal with the awkward `Schema.compile(...).get(...)` dance you have to do to compile the schema. All of that is abstracted away if you pass a `schemaDefinition` instead of a `schema` to `EditorProvider`.

Up until now, `defineSchema` has only supported configuring the names of objects and marks used in the editor as other concepts like fields was initially left out.

However, when observing how PTE behaves when pasting from one editor to another, I realise that it's quite handy if you are required to define the `fields` used by block objects, inline objects and annotations. This way, we can filter out any unknown fields on paste. This is not bulletproof of course since we don't check the field type (yet), but it's a start and it might be good enough for 99% of use cases. Props on the objects that start with `_` are left untouched.

All objects that go into the editor pass through `parseBlockObject(...)` or `parseInlineObject(...)`, which is convenient since we then only have to check the fields in one place.

Notes:

- This should not be a breaking change in neither Studio, Create or Canvas since all these apps pass in a `schema` with fields. However, it could be seen as a breaking change in standalone PTE for those using `defineSchema`. I do however think it's fine to make this change in a minor release.
- Fields on annotations have been left out for now, but can be added if this change of adding fields to inline objects and block objects is deemed sensible.


snorrees (commented):
I take it we would put all the _ stuff from Canvas using this right?

packages/editor/src/internal-utils/parse-blocks.ts
```diff
@@ -60,21 +60,45 @@ function parseBlockObject({
...
   return {
-    ...blockObject,
+    _type: schemaType.name,
     _key: options.refreshKeys
       ? context.keyGenerator()
       : typeof blockObject._key === 'string'
         ? blockObject._key
         : context.keyGenerator(),
+    ...values,
+    ...hiddenValues,
```
snorrees, line 101: Wont you still have the canvas cp bug here? Since you retain _map ect?

packages/editor/src/internal-utils/parse-blocks.ts
```diff
@@ -296,22 +320,43 @@ function parseInlineObject({
...
+    ({name}) => name === inlineObject._type,
+  )
+
+  if (!schemaType) {
     return undefined
   }
 
+  // Validates all props on the inline object and only takes those that match
+  // the name of a field
+  const values = schemaType.fields.reduce<Record<string, unknown>>(
```
snorrees, line 333: Seems like there is some possiblity of DRY'ing between span and block? This and hidden stuff is duplicated 100% isnt it?

christianhg: Good point. I've now created a `parseObject(...)` function where most object parsing is handled.

packages/editor/src/internal-utils/parse-blocks.ts
christianhg: Yes. This PR retains all props starting with `_`.

christianhg: I've updated the PR now so all fields have to be defined, even those for props starting with `_`.

christianhg:
> I take it we would put all the _ stuff from Canvas using this right?

Yes!

PR #1085, opened 2025-04-23
fix: provide `Editor` from a global Context

christianhg:
I guess it doesn't hurt to try this out.

Code more-or-less copy/pasted from https://github.com/sanity-io/ui/blob/main/src/core/lib/createGlobalScopedContext.ts


snorrees (approved):
🤞

snorrees:
Lol

Issue #192, opened 2024-09-17
Rendering list item correctly? 


ryanmearns:
When I attempt to render a list item, I'm unsure how to wrap the list items into an appropriate ordered list. Any guidance how to do this? My code so far is below. I'm currently getting this in the editor:

1. Item 1
1. Item 1
1. Item 1

```typescript
type PortableTextBlock = BlockAnnotationRenderProps["block"];

function Content() {
  const [value, setValue] = useState<PortableTextBlock[] | undefined>([]);

  return (
    <div className="border rounded-lg overflow-hidden">
      <PortableTextEditor
        value={value}
        onChange={(e) => {
          if (e.type === "mutation") {
            const patches = e.patches;
            setValue((currentValue) => {
              return applyAll(currentValue, patches);
            });
          }
        }}
        schemaType={{
          type: "array",
          name: "portable-text",
          of: [
            {
              name: "break",
              type: "object",
              icon: SeparatorHorizontalIcon,
              fields: [
                defineField({
                  name: "style",
                  type: "string",
                  options: {
                    list: ["break"],
                  },
                  validation: (Rule) => Rule.required(),
                }),
              ],
            },
            {
              type: "block",
              name: "block",
              marks: {
                /**
                 * Default decorators:
                 * { "title": "Strong", "value": "strong" },
                 * { "title": "Emphasis", "value": "em" },
                 * { "title": "Code", "value": "code" },
                 * { "title": "Underline", "value": "underline" },
                 * { "title": "Strike", "value": "strike-through" }
                 */
                decorators: [
                  // React.ComponentType
                  {
                    title: "Strong",
                    value: "strong",
                    icon: () => <BoldIcon className="w-4 h-4" />,
                  },
                  // React.ComponentType
                  { title: "Emphasis", value: "em", icon: ItalicIcon },
                  // React.ReactNode
                  {
                    title: "Code",
                    value: "code",
                    icon: <CodeIcon className="w-4 h-4" />,
                  },
                  {
                    title: "Underline",
                    value: "underline",
                    icon: UnderlineIcon,
                  },
                  {
                    title: "Strike",
                    value: "strike-through",
                    icon: StrikethroughIcon,
                  },
                ],
                annotations: [
                  {
                    name: "link",
                    type: "object",
                    icon: LinkIcon,
                  },
                  {
                    name: "comment",
                    type: "object",
                    icon: MessageSquareTextIcon,
                    fields: [
                      {
                        type: "string",
                        name: "text",
                      },
[...]

PR #1093, opened 2025-04-26
fix: Update css for list rendering

rmarscher:
Related to #192.

- Uses 40px left padding (controlled via css variable)
- Uses css transform to scale down and position the bullet symbols
- Right aligns numbers
- Uses a separate counter for number and bullets

The last of these, enables this output list structure to match the browser behavior:

```html
<ol>
  <li>
    Item 1
    <ul>
      <li>Nested bullet</li>
    </ul>
  </li>
  <li>Item 2</li>
</ol>
<ul>
  <li>Level 1 bullet</li>
</ul>
<ol>
  <li>
    Item 1
  </li>
</ol>
```


christianhg:
Hi @rmarscher 

Thanks a lot for this. List rendering is definitely on our radar as something that should not only be improved in the Playground, but also made easier for the general consumers of this library.

I think we'll take an approach which involves adding new data attributes to the DOM to drastically simplify the CSS needed to achieve proper list rendering. You can see a POC draft PR here: https://github.com/portabletext/editor/pull/1145

I liked your idea of `--list-padding` so I copied that to the draft PR. It might be that there are other improvements in here that would benefit the example apps in this repo, but those will have to wait until I'm sure what the final state of #1145 is.

Thank you!

rmarscher:
@christianhg Great. I'm looking forward to it. Thank you. 😄

PR #1095, opened 2025-04-27
feat(behaviors): simplify event propagation by introducing `forward(event)`

christianhg:
Consider this: If someone today asks "how does a Behavior affect event propagation?", then the answer would be:

It depends.

1. If your `guard` is truthy, but you didn't configure any `actions`, then the `event` is forwarded to the next Behavior.
2. If you return any Action from `actions` then your Behavior "owns" the `event.
3. Except if you only do `effect` Actions. Those will not affect the event propagation, because it should be possible to do side effects without interrupting the `event`.
4. However, if you do want to interrupt the `event` while doing the side effect, you can reach for the special `noop` action.

After the introduction of `forward` (and the removal of `noop`), the answer is:

1. If your `guard` is truthy you own the `event`.

It's that simple.

The rest of the rules are easy to learn:

1. If you don't do any `actions`, event propagation stops and nothing happens.
2. If you only do `effect`s then that's that. Nothing happens (except your effect).
3. And you have the option to `execute`, `forward` or `raise` the same `event` or completely different Events if you like.

`execute` and `raise` already exist and with the introduction of `forward` we now have three options on how to granularly control event propagation:

1. You can `execute` Synthetic Events to resolution. This means that executing `insert.text` inserts the text. No Behavior will intercept it.
2. You can `forward` any Event type. This very handy for composition if you want to let the remaining Behaviors with lower priority respond to an `event` without altering it (if you `raise` the same `event` again without altering it then you most likely end up in an infinite loop.)
3. You can `raise` Synthetic and Custom Events, which means they are sent back into the machinery and go through the entire Behavior stack again.

In conclusion, introducing `forward` and removing `noop` simplifies how we think about event propagation. You have to learn the differences between `execute`, `forward` and `raise`, but when those are learned, the rules are simple. And the benefit is clear when looking at the improved `TextFileDeserializerPlugin` and `ImageDeserializerPlugin` in the Playground that no longer have to mutate the `dataTransfer` in order to avoid an infinite loop when raising the `'deserialize'` event again.

### DIFF
--- a/apps/docs/src/content/docs/guides/behavior-cheat-sheet.mdx
+++ b/apps/docs/src/content/docs/guides/behavior-cheat-sheet.mdx
@@ -46,7 +46,7 @@ Read more about using behaviors and building your own with these guides:
 
 ## Log inserted text
 
-Send and `effect` type action along with an `effect` to perform side effects.
+Send and `effect` type action along with a `forward` action to perform side effects without altering the chain of events.
 
 ```js
 const logInsertText = defineBehavior({
@@ -59,12 +59,16 @@ const logInsertText = defineBehavior({
           console.log(event)
         },
       },
+      {
+        type: 'forward',
+        event,
+      },
     ],
   ],
 })
 ```
 
-The `effect` action also has a shorthand function:
+The `effect` and `forward` actions also have shorthand functions:
 
 ```tsx
 const logInsertText = defineBehavior({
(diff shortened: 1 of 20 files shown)


snorrees (approved):
This looks super clean, and makes ALOT of sense! Getting rid of noop alone is a big deal, but forward fits well within the paradigm.

Only thing I would consider, is if there could be some sort of infinite loop detection that short circuits the event loop after _some long number of iterations_, and logs an error.

PR #1145, opened 2025-05-06
feat: add `data-level`/`data-list-index`/`data-list-item` on text blocks

christianhg:


christianhg:
Going to close this and probably go with #1151 instead.

PR #1287, opened 2025-06-15
New example for portable text editor

raptoria:
Adds an example repo for  @portabletext/editor which showcases more "advanced" features such as integration with links and file uploads.


christianhg:
Thank you for this, @raptoria! It's really cool to see how far you've taken the editor in this example and we appreciate the time it must have taken to bring this together.

As you know, we've recently incorporated similar changes in the [Playground](https://playground.portabletext.org/), and I wonder if your advanced example is in fact too similar to the Playground.

I think I would prefer invest in the Playground, and make *that* the go-to example of how to build your own editor with PTE and gradually augment it with plugins to extend functionality.

What do you think?

raptoria:
Yeah agreed, it doesn't make sense for you to have to maintain this as well, especially if you're going to make plugins. Mine is more of a real world example with a lot of external dependencies that is a probably too much overhead and doesn't make sense to be part of the main repo. In any case, I've moved it here: https://github.com/raptoria/portable-text-editor-advanced

PR #1448, opened 2025-07-28
feat: consider range decorations changed if `payload` has changed

juice49:
Previously, range decorations would only be considered changed if their anchor or focus positions had changed. However, there are other parts of the range decoration that may change between renders:

1. The component rendered.
2. The payload data.

For many use cases, it's desirable to re-render the range decoration when these underlying properties change. This commit adds the payload data to the object compared to achieve this. If the developer wishes to ensure the range decoration component is re-rendered, they can do so by adding some unique property to the payload data (such as a key or id).


christianhg (approved):
I don't remember why the `payload` wasn't originally included in the equality check, but it seems reasonable to do so!

I'm interested in the concrete use case though!

PR #1532, opened 2025-08-15
fix: reduce peer deps duplication and set min node 22 version

stipsan:
Node v22 added support for `require(esm)` in 22.12, we've updated the node engines semver in `sanity` and other packages to reflect this, for the same reasons we do it for `22.19`.
While here I noticed the usage of `workspace:*` and `workspace:^` is a bit inconsistent and not as optimized as it could be.
The guideline is that generally `workspace:*` should always be used, as it means the exact version.
Technically, a package like `apps/playground/package.json` isn't published to npm, so what it wants is the equivalent of `workspace:`, it wants the workspace package, whatever the version is, semver ranges doesn't matter. (That might be why the default syntax for `catalog` is `catalog:` and not `catalog:*`).

When packages are published it matters what comes after the `:`. `workspace:*` is replaced with the exact version of the package it refers to. That makes it suitable to `devDependencies`, if `package-a` only has a `devDependency` on `package-b` then it might be useful to see when inspecting `package-a`'s exact version of `package-b` at the time of publish, especially if `package-b` has since published new versions while `package-a` has not.
For `dependencies` and `peerDependencies` the `workspace:^` should be used. That way, if `sanity` bumps `@portabletext/block-tools` but haven't bumped `@portabletext/editor` yet then we don't get two instances of `@portabletext/block-tools` unless it's necessary (they are different major versions).

### DIFF
--- a/apps/playground/package.json
+++ b/apps/playground/package.json
@@ -17,10 +17,10 @@
     "test:unit:watch": "vitest"
   },
   "dependencies": {
-    "@portabletext/block-tools": "workspace:^",
-    "@portabletext/editor": "workspace:^",
+    "@portabletext/block-tools": "workspace:*",
+    "@portabletext/editor": "workspace:*",
     "@portabletext/keyboard-shortcuts": "workspace:*",
-    "@portabletext/patches": "workspace:^",
+    "@portabletext/patches": "workspace:*",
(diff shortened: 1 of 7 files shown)


christianhg (approved):
Thank you so much for this!

PR #1542, opened 2025-08-20
refactor: replace release-please with changesets

stipsan:


packages/patches/LICENSE
stipsan, line 1: pnpm uses the root `LICENSE` file during publishing if the workspace does not have its own file

PR #1574, opened 2025-08-22
feat: support `PortableTextEditable` callback ref

christianhg:
The logic also attempts to simplify the internals of `PortableTextEditable` by
avoiding keeping track of a local `ref` and a local `activeElement`. Instead,
the `MutationObserver` on the DOM node is set up in a callback ref itself, and
`validateSelection` has been pulled out as a standalone function.

### DIFF
--- a/packages/editor/src/editor/Editable.tsx
+++ b/packages/editor/src/editor/Editable.tsx
@@ -5,14 +5,11 @@ import {
   useCallback,
   useContext,
   useEffect,
-  useImperativeHandle,
   useMemo,
-  useRef,
   useState,
   type ClipboardEvent,
   type FocusEventHandler,
   type KeyboardEvent,
-  type MutableRefObject,
   type TextareaHTMLAttributes,
 } from 'react'
 import {Editor, Transforms, type Text} from 'slate'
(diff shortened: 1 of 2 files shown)


packages/editor/src/editor/Editable.tsx
```diff
@@ -1037,13 +934,45 @@ export const PortableTextEditable = forwardRef<
     [onDragLeave, editorActor, slateEditor],
   )
 
+  const callbackRef = useCallback(
+    (node: HTMLDivElement | null) => {
+      if (typeof forwardedRef === 'function') {
+        forwardedRef(node)
+      } else if (forwardedRef) {
+        forwardedRef.current = node
+      }
```
stipsan, line 943: It would be nice if we one day could get rid of the ref forwarding as well (as it's so difficult to optimize and narrow scope if we always have to make everything available to the parent component) 🥲

stipsan (approved):
Such a nice cleanup. THIS IS THE WAY

packages/editor/src/editor/Editable.tsx
christianhg: This would be a breaking change, right? But if it's worth it, it's worth it!

PR #1615, opened 2025-09-01
Enable more React Compiler checks

stipsan:
Uses `eslint.config.js` instead of inline rules, this is mostly due to the increase in rules that the compiler supports since it moved from `eslint-plugin-react-compiler` and became part of `eslint-plugin-react-hooks`.

When doing:
```js
console.log(Object.keys((await import('https://esm.sh/eslint-plugin-react-hooks@latest')).rules))
```
You'll get just two rules: `['rules-of-hooks', 'exhaustive-deps']`. Now, with the compiler added, there's a whole bunch:
```js
console.log(Object.keys((await import('https://esm.sh/eslint-plugin-react-hooks@rc')).rules))
// ['exhaustive-deps', 'rules-of-hooks', 'no-unused-directives', 'hooks', 'capitalized-calls', 'static-components', 'use-memo', 'component-hook-factories', 'preserve-manual-memoization', 'immutability', 'globals', 'refs', 'memoized-effect-dependencies', 'set-state-in-effect', 'no-deriving-state-in-effects', 'error-boundaries', 'purity', 'set-state-in-render', 'invariant', 'todo', 'syntax', 'unsupported-syntax', 'config', 'gating', 'rule-suppression', 'automatic-effect-dependencies', 'fire', 'fbt']
```

By changing the setup to extend `...reactHooksPlugin.configs['flat/recommended']` we'll also enable whatever other checks that are now enabled in the compiler itself. When doing so I spotted a lot of `react-hooks/refs` violations.
One was easy to fix (`useConstant`), the rest is not and is why I set the rule to `warn` instead of the default `error`.

[Take a look at the error shown on `useConstant`:](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEASggIZyEBmMEGBA5DGRfQDoB2HeAnlggQGEI7MHlLs8JSgB4AKkJFiJAPgIBeAsAJxho8XmQF5upYQC+HDggAeOfAQAmCSqSgAbKlHYU0wglDAEBT0JOWDTZQAKDgICFwpcbkNIgEp1VWNFfQAaDhTDTJDCYBjtEwImSnViZjwAOgCEKQJpQRN9ZtbC0wJlVUj2dzcUy3ZYtCrIgEJKurhYJgk0krHYiuc5hYQJaq0dLIlDeLxE1IILVYvSpjxYMdn5mEX6-aKOMxAzIA)
```bash
Found 3 errors:

Error: Cannot access refs during render

React refs are values that are not needed for rendering. Refs should only be accessed outside of render, such as in event handlers or effects. Accessing a ref value (the `current` property) during render can cause your component not to update as expected (https://react.dev/reference/react/useRef)

   8 |   const ref = React.useRef < ConstantRef < TConstant >> (null)
   9 |
> 10 |   if (!ref.current) {
     |        ^^^^^^^^^^^ Cannot access ref value during render
  11 |     ref.current = { constant: factory() }
  12 |   }
  13 |

Error: Cannot access refs during render

React refs are values that are not needed for rendering. Refs should only be accessed outside of render, such as in event handlers or effects. Accessing a ref value (the `current` property) during render can cause your component not to update as expected (https://react.dev/reference/react/useRef)

   9 |
  10 |   if (!ref.current) {
> 11 |     ref.current = { constant: factory() }
     |     ^^^^^^^^^^^ Cannot update ref during render
  12 |   }
  13 |
  14 |   return ref.current.constant

Error: Cannot access refs during render

React refs are values that are not needed for rendering. Refs should only be accessed outside of render, such as in event handlers or effects. Accessing a ref value (the `current` property) during render can cause your component not to update as expected (https://react.dev/reference/react/useRef)

  12 |   }
  13 |
> 14 |   return ref.current.constant
     |          ^^^^^^^^^^^^^^^^^^^^ Cannot access ref value during render
  15 | }
```

[The compiler is a bit strict on `useRef`, due to the issues that might happen if the ref is written to during a concurrent render for example. The good news is that in cases like this, where we want to run `init` exactly once, and then use it for the lifetime of the component, then `useState` is a better candidate.](https://playground.react.dev/#N4Igzg9grgTgxgUxALhASwLYAcIwC4AEASggIZyEBmMEGBA5DGRfQDoB2HeAnlggQGEI7MHlLs8JSgB4AKkJFiJAPgIBeAsAJxho8XmQF5upYQC+HDggAeOfAQAmCSqSgAbKlHYU0wglDAEBT0JOWDTZQAKDgICFwpcbkNIgEp1VWNFfQAaDhTDTJDCYBjtEwIAbR0siQBddWJmPAA6AIQAZTE8BDCTfVVI1PS48jxE1JTLdlimPFhp6qKOMxAzIA)

While at it I've made sure that `typescript` are on the same version in all the workspaces, so that we don't have any issues when v6 comes out 🙌


christianhg:
This is great! I'll take a look at the remaining ref warnings. Maybe I can fix them so we can merge the PR with the rule set to `error`.

christianhg (approved):

Issue #1648, opened 2025-09-12
Block tools image error


silvertech-daniel:
Using `htmlToBlocks()` with the following type errors. It works when the image is removed.

```ts
Schema.compile({
	types: [
		{
			type: 'array',
			of: [
				{
					type: 'block'
				},
				{
					type: 'image',
					name: 'customImage'
				}
			]
		}
	]
})
```

The callstack is (bottom to top)
1. `htmlToBlocks`
2. `sanitySchemaToPortableTextSchema`
3. `portableTextMemberSchemaTypesToSchema`
4. [`Field`'s `type` getter](https://github.com/portabletext/editor/blob/52d931b59bc409a1d607ab06400b85e68b3d1645/packages/sanity-bridge/src/portable-text-member-schema-types-to-schema.ts#L29)
5. [`extendMember`/`ensure`](https://github.com/sanity-io/sanity/blob/bd3d3638612c4c605173390df495036a2a01fab6/packages/%40sanity/schema/src/legacy/Schema.ts#L31C1-L43C4) where it fails to find `sanity.imageHotspot` in the registry

For searchability, the error message is **`Exception has occurred: Error: Unknown type: sanity.imageHotspot`**

christianhg:
Can you give me a complete snippet that would fail? Perhaps even as a failing test case for `block-tools`.

delucis:
I hit this same error following https://www.sanity.io/learn/course/migrating-content-from-wordpress-to-sanity/converting-html-to-portable-text#s-125c13b1f1fa

This useses `const defaultSchema = Schema.compile({types: schemaTypes})` where `schemaTypes` is the user-defined Sanity schema definitions.

Updating it as suggested above to use `builtinTypes` resolved the `Unknown type: sanity.imageHotspot` error:

```js
const defaultSchema = Schema.compile({types: [...schemaTypes, ...builtinTypes]})
```

Dependency versions:

```json
"@portabletext/block-tools": "^4.1.1",
"@sanity/schema": "^4.18.0",
```

PR #1665, opened 2025-09-19
test: add unit tests for custom image schema handling in htmlToBlocks

AryanJ-NYC:
I added comprehensive test coverage for the schema configuration reported in issue #1648. The test validates that `Schema.compile()` incorrectly handles an array type containing both block and customImage types when used with htmlToBlocks().

##  What was implemented
Created custom-image-schema.test.ts in /packages/block-tools/test/html-to-blocks/ with four test cases:

1. Basic schema with block and customImage types - Validates that a schema with an array containing both type: 'block' and type: 'image' with name: 'customImage' compiles correctly and converts HTML to blocks without errors.
2. Mixed content conversion - Tests that HTML containing both text paragraphs and image elements properly converts to block structures when using the custom schema.
3. Text formatting with customImage in schema - Ensures that text blocks with formatting (bold, italic) work correctly even when the schema includes the customImage type definition.
4. Complex content structures - Validates that nested content like blockquotes converts properly with the customImage-enabled schema.

## Key aspects

  - The test uses the exact schema structure from the issue: an array with `{ type: 'block' }` and `{ type: 'image', name: 'customImage' }`
  - Follows existing test conventions in the codebase (using vitest, JSDOM, and the test key generator)
  - Validates that `htmlToBlocks()` works without throwing "Unknown type" errors
  - Ensures the schema compilation succeeds and produces the expected block output

This test serves as both validation that the reported issue can be handled correctly and as regression prevention for future changes to schema compilation or HTML conversion logic.

### DIFF
--- a/packages/block-tools/test/html-to-blocks/custom-image-schema.test.ts
+++ b/packages/block-tools/test/html-to-blocks/custom-image-schema.test.ts
@@ -1,0 +1,279 @@
+import {Schema} from '@sanity/schema'
+import {JSDOM} from 'jsdom'
+import {describe, expect, test} from 'vitest'
+import {htmlToBlocks} from '../../src'
+import {createTestKeyGenerator} from '../test-key-generator'
+
+describe('custom image schema', () => {
+  test('Schema with array of block and customImage', () => {
+    const customSchema = Schema.compile({
+      name: 'withCustomImage',
+      types: [
+        {
+          type: 'object',
+          name: 'document',
+          fields: [
+            {
+              title: 'Content',
+              name: 'content',
+              type: 'array',
+              of: [
+                {
+                  type: 'block',
+                },
+                {
+                  type: 'image',
+                  name: 'customImage',
+                },
+              ],
+            },
+          ],
+        },
+      ],
+    })
+
+    const blockContentType = customSchema
+      .get('document')
+      .fields.find((field: any) => field.name === 'content').type
+
+    const result = htmlToBlocks(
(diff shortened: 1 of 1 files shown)


christianhg:
Thank you for this, although I must be honest and say this PR confuses me a bit.

I starts out by stating:

> The test validates that **`Schema.compile()` incorrectly handles an array type containing both block and customImage** types when used with htmlToBlocks().

Later, it reads:

> Basic schema with block and customImage types - **Validates that a schema with an array containing both type: 'block' and type: 'image' with name: 'customImage' compiles correctly** and converts HTML to blocks without errors.

In any case, all the test cases added fail because `Schema.compile` fails with `Exception has occurred: Error: Unknown type: sanity.imageHotspot`. As already stated in #1647, you can fix this by spreading the `builtinTypes` into the `types` array. To fix the compilation in the first test case you also need to rename the object to something other than `document`, for example `article`:

```ts
import {builtinTypes} from '@sanity/schema/_internal' // <-- import this

const customSchema = Schema.compile({
  name: 'withCustomImage',
  types: [
    {
      type: 'object',
      name: 'document', // <-- don't use `document` as object name
      fields: [
        {
          title: 'Body',
          name: 'body',
          type: 'array',
          of: [
            {
              type: 'block',
            },
            {
              type: 'image',
              name: 'customImage',
            },
          ],
        },
      ],
    },
    ...builtinTypes, // <-- and use it here
  ],
})
```

I'll close this PR. The goal of the PR requested in #1648 should be to prove that you **cannot** use `block-tools` with a schema containing an `image` type, but this PR clearly shows that it's possible.

PR #1687, opened 2025-09-24
fix: avoid render functions causing unmounts

christianhg:
By using a stable wrapper component that isn't being recreated, this change keeps the React Compiler happy and mitigates re-render issues that can affect downstream usages like in Sanity Studio.

The previous `<props.renderX` trick was used to avoid the React Compiler flagging Ref violations when calling `props.renderX(` since that call references Ref Objects during render.

However, `<props.renderX` causes unnecessary unmounts because the React element is being recreated over and over. To mitigate this, and keep the React Compiler happy, `props.renderX(` is now called in a wrapper component instead.

### DIFF
--- a/.changeset/slick-apples-post.md
+++ b/.changeset/slick-apples-post.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: avoid render functions causing unmounts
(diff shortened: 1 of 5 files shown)


stipsan (approved):
What a mystery this one was 😮‍💨

PR #1799, opened 2025-11-01
fix: remove `get-random-values-esm`

chuttam:
The `get-random-values-esm` dependency is out-of-date and now includes an [upstream vulnerabilty](https://github.com/kenany/get-random-values/issues/293).

Although [an open PR](https://github.com/sanity-io/get-random-values-esm/pull/14) exists to address the dependency issue, this commit prefers to replace the mechanism of `getRandomValues` with native functionality rather than the external dependency.

This project already requires node v20+, which allows the safe use of `globalThis.crypto`.

### DIFF
--- a/packages/block-tools/src/util/randomKey.ts
+++ b/packages/block-tools/src/util/randomKey.ts
@@ -2,9 +2,10 @@ export function keyGenerator() {
   return randomKey(12)
 }
 
-function cryptoRNG(length = 16) {
+// WHATWG crypto RNG - https://w3c.github.io/webcrypto/Overview.html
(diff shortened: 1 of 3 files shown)


christianhg:
> @christianhg I'm not sure if there's anything I need to do for the one failing flavour of Playwright test? Doesn't seem to be a way for me to re-trigger to see if it's a code problem or environmental problem.

Hi @chuttam. A few of the tests are a bit flaky, don't worry!

christianhg (approved):
Thank you for this! Always nice to cut dependencies.

I just minimized the diff a bit and added changesets.

PR #1803, opened 2025-11-03
fix: issues with expanding selection down using Shift+ArrowDown

christianhg:


christianhg:
Closing as stale: the branch predates the refactors that rewrote the keyboard behaviors and `editor-dom.ts` it touches. If Shift+ArrowDown selection expansion still misbehaves, a fresh red test against current `main` is a cheaper restart than rebasing this.

PR #1964, opened 2025-12-05
chore: add 'better rendering' RFC

christianhg:


packages/editor/RFC-renderer-registration.md
```diff
@@ -0,0 +1,216 @@
...
+
+### Phase 4: Testing
+
+- Test with existing plugins
+- Verify no breaking changes to render props
+- Performance testing
+
+## Open Questions
+
+- Should we provide convenience hooks like `useNodeFocused(nodeKey)` and `useNodeSelected(nodeKey)`?
```
pepahlavacek: ohh, that would be neat!

packages/editor/RFC-renderer-registration.md
```diff
@@ -0,0 +1,216 @@
...
+editor.registerRenderer<typeof schema>({
+  type: 'decorator',
+  name: 'highlight',
+  render: (props) => <HighlightDecorator {...props} />,
+})
+```
+
+## Resolution Order
+
+1. Registered renderers (by priority, highest first)
```
pepahlavacek: how is priority defined? (order of registration? order in the renderer plugin? first defined executes first?

christianhg: Order of registration and then the first defined wins. Like Behaviors. (Until we add a Priority API which we would want for Behaviors as well at one point)

PR #1975, opened 2025-12-08
fix(deps): bundle Slate packages to prevent version conflicts

christianhg:
Slate packages (`slate`, `slate-dom`, `slate-react`) are now bundled into the
`@portabletext/editor` output instead of being external dependencies.

This prevents issues when multiple versions of `@portabletext/editor`
exist in the same application. For example, Canvas depends on both
`@portabletext/editor` directly and on Sanity Studio, which depends on its
own version of `@portabletext/editor`. Since `slate-react` declares
`slate-dom` as a peer dependency, each PTE version brought its own
`slate-dom` instance, causing WeakMap and state conflicts that broke the
editor.

By bundling Slate:

1. Each PTE version has its own isolated Slate instance
2. No shared `slate-dom` to conflict over
3. Consumers don't need to manage Slate version overrides

This is shouldn't be a breaking change since Slate is an implementation detail
and not exposed in the public API.

Note: The API Extractor had to be disabled since it errors out when bundling
Slate. Slate is indirectly exposed in public APIs which is unfortunate, but
very hard to fix. We'll therefore miss out in API Extractors tsdoc validation,
which feels like a reasonable tradeoff.

### DIFF
--- a/.changeset/bundle-slate-packages.md
+++ b/.changeset/bundle-slate-packages.md
@@ -1,0 +1,8 @@
+---
+"@portabletext/editor": patch
+---
+
+Bundle Slate packages to prevent version conflicts
+
+Slate packages (`slate`, `slate-dom`, `slate-react`) are now bundled into the `@portabletext/editor` output instead of being external dependencies. This prevents issues when multiple versions of `@portabletext/editor` exist in the same application, where shared `slate-dom` instances could cause WeakMap and state conflicts.
+
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -28,7 +28,15 @@
     {
       "matchFileNames": ["packages/editor/package.json"],
       "matchDepTypes": ["devDependencies"],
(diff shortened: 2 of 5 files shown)


christianhg:
> Disabling `extract` means tsdoc tags are no longer checked with `@microsoft/api-extractor` under the hood, that's probably fine?

This is what Cursor suggested for the longest time, but I had hoped for a better solve. Oh well, maybe it's fine, as you say.

> Moving deps from `dependencies` into `devDependencies` will bundle them so you don't need to specify `extract.bundledDependencies` or `external`.

Nice! I was actually unsure about this! I know tsup works that way, but I remember pkg-utils didn't always do this. Or maybe I'm misremembering 🙂 

> Struggling with parsing can indicate that even though `slate` is supposed to be an internal implementation detail it's exposed in the public API someway that leads to it needing to do a bunch of inference which in turn leads it to process a bunch of slate dts.

Exactly! Yes, it's most likely exposed in some indirect way that is very hard to fix.

Thank you for your input, it works now 🙇

PR #2009, opened 2025-12-10
fix: isolate and limit `@sanity/types` usage

christianhg:
We want to decouple PTE and friends from `@sanity/*`. This will require breaking changes, but there are some non-breaking changes we can do in the meantime:

1. All runtime usages of `@sanity/types` have been replaced with local equivalents or equivalents from `@portabletext/schema`: https://github.com/portabletext/editor/pull/2015
2. All types that are already mirrored and live in `@portabletext/schema` are now imported from there instead.
3. `PortableTextChild` and `PortableTextListBlock` now also exist in `@portabletext/schema` so we can use those instead.
4. `Path` and `KeyedSegment` have been copy/pasted to the `editor` package.

End result:

PTE now only imports and exports the following from `@sanity/types` and the imports have been isolated to a single file while Biome is used to prevent accidental imports from `@sanity/types` in the future:

```ts 
  ArrayDefinition,
  ArraySchemaType,
  BlockDecoratorDefinition,
  BlockListDefinition,
  BlockStyleDefinition,
  ObjectSchemaType
```

### DIFF
--- a/packages/editor/biome.json
+++ b/packages/editor/biome.json
@@ -1,0 +1,17 @@
+{
+  "extends": "//",
+  "linter": {
+    "rules": {
+      "style": {
+        "noRestrictedImports": {
+          "level": "error",
+          "options": {
+            "paths": {
+              "@sanity/types": "Import from 'types/sanity-types' instead to maintain visibility over @sanity/types usage."
+            }
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


stipsan (commented):
Forgive me if this is not relevant but is there any potential impact here to how `IntrinsicDefinitions` works in the way `sanity` implements it, that rely on `@sanity/types` not being inlined?
If so maybe we should use a changeset prerelease here to test the changes on a draft PR at https://github.com/sanity-io/sanity first?
If there's no potential impact I say we merge and release this 🥳

PR #2334, opened 2026-03-06
fix: merge consecutive text ops when one undoStepId is undefined

christianhg:
Companion to #2243. That PR fixed the case where both undoStepIds are defined but different (consecutive forwarded events). This PR fixes the remaining case where the current operation has no undoStepId but the previous one did.

This happens when a forward-only behavior intercepts one keystroke (giving it an undoStepId from the send entry) but the next keystroke has no matching behavior (so its undoStepId is undefined). The mismatch prevented the consecutive insert_text merge heuristic from running, making each keystroke its own undo step.

The reverse case (current has ID, previous doesn't) is intentionally not merged. When a behavior claims an event and gets an undoStepId, that signals an intentional undo step boundary, like input rules replacing `->` with `→`.

### DIFF
--- a/.changeset/fix-undo-step-id-leak-undefined.md
+++ b/.changeset/fix-undo-step-id-leak-undefined.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: merge consecutive text ops when one undoStepId is undefined
--- a/packages/editor/src/editor/undo-step.ts
+++ b/packages/editor/src/editor/undo-step.ts
@@ -117,6 +117,40 @@ export function createUndoSteps({
     }
   }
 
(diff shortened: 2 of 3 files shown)


christianhg:
Picks up the work from #2334 — closing in favor of this. The underlying gap still existed after the renames in #2700 and the `createUndoSteps` reorder in #2739, so a fresh PR on top of current main was cleaner than a rebase. Same fix shape; updated `Path.equals` → `pathEquals`, adapted to the new `createUndoSteps` signature, and reshaped the test to assert via `history.undo` semantics (one undo restores empty value) rather than introspecting the stack directly — `userEvent.type` adds an initial focus-selection step that would have skewed a `toHaveLength(N)` check.

PR #2364, opened 2026-03-12
fix: resolve block path correctly when dropping block on inline object

christianhg:
When a schema allows the same type name (e.g. `image`) as both a block object and an inline object, dropping a block image onto an inline object turns it into an inline image instead of inserting it as a block. The root cause is in `operation.insert.block.ts` where the start/end block resolution uses `path.length <= start.path.length`, which can match inline nodes (path length 2) when the DnD handler has already moved the selection inside a text block at the inline object's position. The `insert.block` operation then inserts at the inline object's path rather than the containing block's path.

The fix changes both start and end block resolution to `path.length === 1`, ensuring `insert.block` always resolves top-level blocks. This is consistent with how the rest of the operation already treats these paths (e.g. `endBlockPath[0]!`).

The test sets up a schema with `image` in both `blockObjects` and `inlineObjects`, creates a document with a text block containing an inline image and a separate block image, then drops the block image onto the inline object and verifies it remains a top-level block.

### DIFF
--- a/.changeset/fix-dnd-block-on-inline.md
+++ b/.changeset/fix-dnd-block-on-inline.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: resolve block path correctly when dropping block on inline object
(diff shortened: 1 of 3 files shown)


christianhg:
Closing: the bug no longer reproduces. This PR's own regression test, applied unchanged onto today's `main` (`8b85bb5bc`) without the fix, passes on chromium and firefox — the block image stays a top-level block when dropped onto an inline object. Evidently fixed by the operation refactors since March. The test lands separately as a regression pin.

PR #2409, opened 2026-03-24
fix: stable span key

christianhg:
Span `_key`s carry identity downstream (Studio diff, content-lake patches, presence cursors). Typing right next to a marked span shouldn't change that identity, but it does today — the `insert.text` core behavior always raises `insert.child`, even when the resulting span would be identical to the one already there.

This PR makes the `insert.text` behavior stand down in two cases:

1. When the active marks already match the focus span's marks, the behavior returns `false` and the engine inserts the character into the existing span. Typing directly after an annotation (caret at offset 0 of the unmarked span that follows) keeps the unmarked span's `_key`.

2. When the caret is at the end of a span whose marks differ from the active marks, but the next span's marks match the active marks, the behavior moves the caret to offset 0 of that next span and re-raises the `insert.text`. Typing with a decorator activated, just before a span that already has that decorator, absorbs into the existing decorated span and keeps its `_key`.

Both scenarios are covered by `tests/stable-keys.test.tsx`.

---

> [!NOTE]
> **Medium Risk**
> Touches core text-insert behavior and mark comparison at span boundaries; wrong guard logic could change where characters land or how marks apply, though scope is narrow and well-covered by tests.
> 
> **Overview**
> **Fixes span `_key` churn** when typing next to annotated or decorated text. The core `insert.text` behavior no longer always raises `insert.child` when active marks would produce the same span as one already in the document.
> 
> When **active marks match the focus span**, the behavior returns `false` so the default engine path inserts into that span (e.g. typing at the start of the unmarked span after a link keeps `bazKey`).
> 
> When the caret is **at the end of a span** whose marks differ from the active set but the **next sibling is a span with matching marks**, it selects offset 0 on that next span and re-raises `insert.text` so text merges into the existing decorated span instead of creating a throwaway span. **Non-span siblings** (e.g. inline objects) are ignored so typing does not jump past them.
> 
> Adds **`isEqualMarkSet`** (order-insensitive mark comparison, distinct from order-sensitive `isEqualMarks`) plus unit and browser tests in `stable-keys.test.tsx`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 55d7a9b2eb287e77b5edf68d4f58aa1223462170. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/cyan-shoes-try.md
+++ b/.changeset/cyan-shoes-try.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: stabilize span keys when writing between marks
--- a/.changeset/cyan-shoes-try.md
+++ b/.changeset/cyan-shoes-try.md
@@ -1,0 +1,14 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: stabilize span keys when writing between marks
+
+Typing next to decorated or annotated text no longer replaces the span
+the character lands in. Previously, when the active marks matched the
+span at the caret (or the span right after it, when typing at a span
+boundary), the editor inserted a fresh span with a fresh `_key` and
+merged it away; anything keyed on span identity (diff views,
+content-lake patches, presence and comment anchors) saw a span vanish
+and an unfamiliar one appear. The character now flows into the existing
+span, and its `_key` is preserved.
--- a/packages/editor/src/behaviors/behavior.core.insert.ts
+++ b/packages/editor/src/behaviors/behavior.core.insert.ts
@@ -1,3 +1,5 @@
+import {isEqualMarks} from '../internal-utils/equality'
+import {getNextSpan} from '../selectors'
 import {getActiveAnnotationsMarks} from '../selectors/selector.get-active-annotation-marks'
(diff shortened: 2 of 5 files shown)


christianhg:
Closing after a deep revival attempt. The bug is alive (the branch's tests are red against current `main`), but the fix belongs to a different layer. Typing after an annotation resolves the caret to the annotated span's end, so keeping the neighbor's key requires rerouting the insert, and every rerouting mechanism is observable in a way that breaks legitimate plugin behavior: a synthetic `select` before the insert resets typing-continuity plugins (emoji picker, red on this branch's last push), `execute` hides the insertion from the behavior chain, and primitive-form `raise` is choreography on alpha surface. The durable rule is at the merge normalization: a merge between a newly-minted span and a pre-existing span keeps the pre-existing key, which needs the provenance the corrections store (#3043) is building toward. Tracked in Linear with the full analysis. The branch preserves `isEqualMarkSet`, the pinning scenarios (annotation typing, absorb, inline-object adjacency), and the mechanism experiments.

PR #2458, opened 2026-04-03
feat: per-style schema restrictions

rexxars:
This one might need a bit of discussion, but here goes:

**Goal:** Allow style definitions in the PTE schema to restrict which decorators, annotations, lists, and inline objects are allowed per block style, so the editor core (toolbar, paste, keyboard shortcuts, normalization) automatically respects these restrictions. This is to for instance disallow links in titles, but allow it in regular text paragraphs.

**Architecture:** Extend `StyleDefinition` and `StyleSchemaType` with optional decorators, annotations, lists, and inlineObjects override arrays. Add a `getStyleFeatures(schema, styleName)` helper that returns the effective feature set for a style (the style’s overrides if present, else the top-level schema arrays). Update all schema consumption points to use this helper when the block style is known. The toolbar already has disable/enable infrastructure - wire the availability listener to check per-style restrictions.

### DIFF
--- a/packages/editor/gherkin-spec/per-style-restrictions.feature
+++ b/packages/editor/gherkin-spec/per-style-restrictions.feature
@@ -42,3 +42,37 @@ Feature: Per-style schema restrictions
     And "h2" is toggled
     And "strong" is toggled
     Then "hello world" has no marks
+
+  Scenario: Annotation toggle is blocked on restricted style
+    Given the text "hello world"
+    When "hello world" is selected
+    And "h1" is toggled
+    And "link" is toggled
+    Then "hello world" has no marks
+
+  Scenario: Annotation toggle works on unrestricted style
+    Given the text "hello world"
+    When "hello world" is selected
+    And "link" "a1" is toggled
+    Then "hello world" has marks "a1"
+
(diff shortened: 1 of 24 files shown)


christianhg:
This looks interesting and relevant and aligns with some of the thoughts I'm having myself:

I'm working on first-class nesting where text `'block'`s can be children of other blocks to thereby create nested, editable structures. For example, a `'code'` block might have an array field that accepts `of: [{type: 'block'}]`.

What I have been thinking about in that regard is how to scope and inherit decorators, annotations, lists and styles, and this PR basically mirrors my thinking:

1. Explicitly configuring them on a nested block overrides them
2. Leaving them `undefined` inherits from the root schema

This means that for a text block in a code block you want to set them all to `[]` while a text block in a callout might want to inherit annotations and decorators from the root.

One thing I've been thinking about in addition, as part of this work, is if it should be possible to define different types of text blocks to thereby scope schema definitions, even at the editor root. But even if that becomes a thing, top-level per-style schema restrictions feel valuable on their own, solve a real problem without inventing a new concept and work out-of-the box with serializers.

PR #2480, opened 2026-04-08
fix: preserve selection after remote mark changes

christianhg:
When a remote user marks text (e.g. bold on "bar" in "foobarbaz"), the block's children are replaced with new spans that have new keys. The sync machine detects this as a `!hasSharedKeys` case and replaces the entire children array atomically via `set_node`.

The problem: `transformPoint` for `set_node` only handles key substitution and text collapse. When the children array is replaced entirely, the old span keys disappear and the selection can't be mapped to the new structure. The cursor jumps to the wrong position.

The fix converts selection points to block-level text offsets before the atomic replacement, then converts back to span-level selection points using the new children. This uses the existing `spanSelectionPointToBlockOffset` and `blockOffsetToSpanSelectionPoint` utilities.

### DIFF
--- a/.changeset/fix-selection-after-remote-mark.md
+++ b/.changeset/fix-selection-after-remote-mark.md
@@ -1,0 +1,5 @@
+---
+"@portabletext/editor": patch
+---
+
+fix: preserve selection after remote mark changes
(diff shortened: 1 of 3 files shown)


christianhg:
Superseded by #2933, which moves the same block-text-offset mapping into `transformPoint`'s `set` case with the text-preservation validity condition enforced where the mapping happens, covering every point consumer at every children-replacement path. This PR's reproduction scenario is ported into #2933's browser suite.

PR #2532, opened 2026-04-22
Container awareness

christianhg:
# Container awareness

The editor has had a concept of "editable containers" since scope-typed schemas landed: nodes that declare a child array field to hold nested blocks. But until this PR the rest of the editor didn't really know about them. Selectors assumed the focused block was at the root. Behaviors hardcoded root-level ancestor walks. Operations resolved keys by index into `editor.children`. The parser copied container field values raw without validating nested content. Renderers looked up styles and decorators on the root schema, ignoring any sub-schema the container declared.

The result: typing inside a code-block could produce orphan DOM. Backspace at the start of a cell deleted the whole table. Pasting blocks into a callout merged them into the surrounding text block. Clicking a span inside a container snapped the cursor to offset 0. Arrow keys escaped `<table>` elements entirely.

This PR makes every observer and mutator in the editor walk into containers at any depth, and adds a scoped `defineLeaf` API so consumers can control the outermost element of spans, inline objects, and void block objects at specific depths.

## Breaking changes

- `BlockPath`, `ChildPath`, and `AnnotationPath` are now aliases for `Path`. Consumers only see a break if they explicitly annotate a local variable with one of those names, or destructure a path as a fixed-length tuple. This cascades into event types that accepted the narrow aliases (`block.set`, `block.unset`, `child.set`, `child.unset`, `delete.block`, `delete.child`, `annotation.set`, `move.block`, `move.block up`, `move.block down`, `select.block`).
- Seven `@public` utility functions that took a context `Pick` now require `containers`: `blockOffsetToBlockSelectionPoint`, `blockOffsetToSpanSelectionPoint`, `blockOffsetToSelectionPoint`, `blockOffsetsToSelection`, `childSelectionPointToBlockOffset`, `selectionPointToBlockOffset`, `sliceBlocks`. Consumers that pass `snapshot.context` (or spread it) are unaffected; standalone context construction needs to add `containers`.

`EventPosition` gains an additive `isContainer: boolean` field next to the existing `isEditor: boolean`. The two are mutually exclusive: `isEditor` is true when the event originated at the editor root, `isContainer` is true when it originated at an editable container wrapper (a code block, a table cell, a callout). Existing consumers reading `isEditor` continue to work.


christianhg:
Closing — `pte-container-awareness` was the source mega-branch; its commits have been split + landed across PR #2536, PR #2542, PR #2543, PR #2544, PR #2545, PR #2546, with the remaining work in PR #2549 (`feat: GA containers`).

PR #2538, opened 2026-04-24
feat: make the engine container-aware

christianhg:
Builds on top of #2536. The editor's selectors, behaviors, operations, parser, normalization, and renderers still assumed content lives at the root of the document. That assumption blocked anything useful inside editable containers: a code block's `lines`, a table cell's `content`, a callout's `block`.

This PR makes every layer work at any depth. `defineContainer` and `defineLeaf` become `@alpha` for consumers to register containers and wire leaf-specific rendering. The playground gains callout and table examples to exercise the full stack end-to-end.

The changes cascade through four areas. Traversal and selectors read schema sub-schemas at the enclosing container, so decorator, annotation, list, and style lookups scope correctly. Behaviors, operations, and the parser/normalization layer accept paths at any depth, with dedicated follow-ups for cross-container range delete, empty-container deletion on Backspace/Delete, nested style handling, and container-escape on trailing empty-line Enter. Renderers gain a `defineLeaf` pipeline and emit minimal DOM inside containers (`data-block-type="container"` on wrappers, no legacy `data-slate-*` noise).

PR #2540, opened 2026-04-24
feat: make the editor container-aware (selectors, behaviors, operations, parser, normalization)

christianhg:
Engine-layer follow-up to #2536. Makes selectors, behaviors, operations, parser, and normalization container-aware so every part of the editor that reasons about the tree uses the full path, not just root-level keys.

Rendering, DOM output, and the `defineLeaf` API stay on top of #2538 — this PR is internals only.

### What changes

- **Selectors, converters, snapshot utilities** compose `getNode` / `getNodes` / `getAncestors` instead of hardcoding root-level walks, so they resolve at any depth.
- **Core behaviors** (insert-break, delete, list, style, decorator, annotation) scope their schema lookups to the enclosing container's sub-schema. Inside a container the sub-schema is authoritative; at root the permissive top-level contract is preserved.
- **Operations** (insert.block, delete, etc.) resolve selection at any depth and handle range delete that crosses container parents.
- **Parser and normalization** walk nested type declarations so blocks registered inside a container are validated against the inner schema, not silently dropped.
- **Enter-escape** behavior: when the caret is in the last block of a container, that block is empty, and its previous sibling is also empty, pressing Enter deletes both trailing empties and inserts a new text block after the outermost container that can host it.
- **Availability selectors** exposed for external plugins: `isAvailableDecorator`, `isAvailableList`, `getAvailableStyles`, `getAvailableLists`, `getAvailableDecorators`. Primary consumer: `plugin-markdown-shortcuts`.

### What this doesn't ship

Rendering and DOM stay unchanged. Containers still render through the legacy pipeline with existing attributes. The `defineLeaf` API, `data-block-type="container"`, minimal container DOM, and playground examples land in a follow-up.


christianhg:
Subsumed by PR #2549 (`feat: GA containers`) which builds on this branch's tip and adds the rendering, DOM, `defineLeaf` API, and playground stress tests on top.

PR #2541, opened 2026-04-24
feat: container-aware rendering, defineLeaf, and minimal container DOM

christianhg:
Rendering layer for container-awareness. Builds on top of #2540 (engine internals). Makes renderers container-aware, finalizes the `defineLeaf` API, and gives registered editable containers a minimal DOM surface.

### What changes

- **`defineLeaf`** API exposed `@alpha` via `@portabletext/editor` (type) and `@portabletext/editor/plugins` (plugin). Registers a custom renderer for a span, inline object, or void block object matched by scope. Same progressive-specificity rules as `defineContainer`. `defineLeaf` on spans is supported — the consumer owns the outermost span element and receives children already wrapped with `renderDecorator` / `renderAnnotation`.
- **Container wrappers** emit `data-block-type="container"` instead of `data-slate-node="element"`. The engine's `toSlatePoint` DOM→model queries are widened to accept both.
- **Minimal DOM inside containers.** Text blocks, spans, and children inside a registered editable container drop the legacy `pt-*` classes, div-soup wrappers, `data-child-key`, `data-child-name`, `data-slate-node="text"`. Kept: `data-block-type`, `data-block-key`, `data-block-name`, `data-child-type`, `data-slate-leaf`, `data-slate-string` (engine-required for mark-split leaf DOM-to-model mapping). Consumers registering `defineContainer` own the render tree underneath.
- **`data-slate-length`** dropped — unused everywhere.
- **Playground** ships callout and table container examples demonstrating `defineContainer` (including `$..callout.block` for per-text-block semantic rendering).

### What this doesn't change

Root-level DOM is unchanged. `data-slate-*` attributes outside containers stay intact for backwards compatibility with consumer CSS and DOM-walking code.


christianhg:
Closing — content subsumed by PR #2549 (`feat: GA containers`).

PR #2544, opened 2026-04-27
fix: support range delete across different container parents

christianhg:
Selecting text from inside an editable container (a code-block line, a callout's text block, a table cell) out to a sibling block and pressing Backspace would throw, because Slate's range-delete assumes the two end blocks are siblings at the same level.

The engine now detects when the selection crosses a structural boundary and applies a sensible default: the partial content at each end is removed, blocks fully contained between the two ends are removed, and the remaining shells are left in place without merging. Same-parent range-deletes still merge as before.

Cherry-picked out of the larger container-awareness branch so it can land independently.

### DIFF
--- a/.changeset/cross-container-range-delete.md
+++ b/.changeset/cross-container-range-delete.md
@@ -1,0 +1,5 @@
+---
+"@portabletext/editor": patch
+---
+
+fix: support range delete across different container parents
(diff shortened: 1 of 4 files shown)


christianhg:
Folded into #2543, which now ships a single canonical `deleteRange` primitive used by both `operation.delete` and `operation.insert.block`. The cross-parent algorithm and tests from this PR are preserved verbatim there.

PR #2547, opened 2026-04-27
fix: container behavior corrections

christianhg:
Inside an editable container, several behaviors were quietly making the wrong call. Markdown shortcuts kept consuming Backspace even when the container scope didn't allow the default style. The decorator/annotation/list/style guards used the root schema instead of the block's resolved sub-schema. Backspace and Delete on a container's last empty line did nothing. Enter on a trailing empty line couldn't escape the container.

This PR brings six corrections together so containers behave the way users expect:

- Decorator, annotation, list, and style lookups in core behaviors now resolve through the block's sub-schema instead of the root schema.
- Nested block styles in containers no longer get `normal` prepended automatically — the container's own list of styles is authoritative.
- Backspace or Delete in an editable container that's already empty deletes the container itself instead of no-op'ing.
- Pressing Enter on a trailing empty line escapes the container, with the empty trailing line removed and a new sibling block inserted after.
- Availability selectors (`getAvailableStyles`, `getAvailableLists`, `getAvailableDecorators`, `getAvailableAnnotations`, `isAvailableList`, `isAvailableDecorator`) are exposed @beta so external plugins can gate behaviors on the focus block's sub-schema.
- The plugin-markdown-shortcuts `clearStyleOnBackspace` no longer consumes the keystroke when the default style isn't available in scope.


christianhg:
Closing — folding scope into PR #2532 as the GA containers PR. All 6 commits here will be absorbed into the rebased pca-ref.

PR #2553, opened 2026-04-27
feat: GA containers

christianhg:
Containers, end-to-end.

`defineContainer` and `ContainerPlugin` have been on the editor for a few releases as `@alpha`, but most of the editor only honored them at the surface. The engine, behaviors, and rendering still acted as if every block lived at the editor root. Registering a callout, code block, or table mostly worked for shallow content but broke down as soon as you started moving the cursor, deleting blocks, applying decorators, or pasting nested structures.

This PR makes containers actually work end-to-end:

- **Selectors and traversal** resolve blocks at any depth, so behaviors that compose them (active marks, available decorators, focused block, sibling navigation) read the right block when the caret is inside a container.
- **Operations** (`block.set`, `block.unset`, `child.set`, `decorator.add/remove`, `annotation.add/remove`, `delete`, `insert.block`, `insert.child`, `move.block`) accept paths at any depth and validate against each block's sub-schema.
- **Parsing and normalization** walk into container fields and validate against the per-level sub-schema, so nested content is not silently rewritten to the root schema's shape.
- **Behaviors** for keyboard, breaks, clicks, and selection navigate inside containers correctly; ArrowUp/Down at container edges no longer let the browser drop the caret outside the editor.
- **Empty-container deletion** removes the container itself when the user hits Backspace or Delete in an empty inner block, walking up through any vacuous ancestor up to the first one with structural siblings.
- **Enter on a trailing empty line** with an empty previous block escapes the container (the familiar "double-Enter" affordance from rich-text editors elsewhere).

On top of the engine work, this also lands the public `defineLeaf` API alongside `defineContainer`. Consumers register `defineLeaf({scope, render})` for spans, inline objects, and void block objects to take full control over the outermost element rendered for matching nodes. The most specific scope wins per JSONPath specificity. Both APIs are `@alpha`.

Container wrappers now emit `data-block-type="container"` instead of `data-slate-node="element"`, and the DOM inside a registered container is minimal: no inner text-block wrappers, no `pt-*` classes, no internal `data-child-*`. Engine queries that resolve DOM events to model paths accept the new attribute set.

Five availability selectors (`getAvailableStyles`, `getAvailableLists`, `getAvailableDecorators`, `isAvailableDecorator`, `isAvailableList`) graduate to `@beta` for plugins that need to gate behavior on what the focused block's sub-schema allows. The markdown-shortcuts plugin is updated to use them.

The playground gains callout, table, and code-block container examples that exercise list rendering, the `$..callout.block` second-container pattern, and per-leaf rendering inside containers.

## Breaking changes

None. All public APIs preserve their shapes; the changes are additive (new `@alpha`/`@beta` exports) or correctness fixes inside `@alpha` machinery.


christianhg:
Superseded by #2568.

Everything else from this branch (cross-parent range delete, delete-empty-container, unwrap-object-containers-only, unhang-range-container-aware, is-overlapping-selection-touching, unify-delete, FEFF Android fix, editor-as-Subscribable) has landed on `main` as separate PRs (#2557, #2558, #2563, #2564, #2566, #2570, #2571, #2574, #2576, #2578, #2584). The remaining `defineLeaf` engine + playground archetypes + container regression tests are now in #2568 (rebased onto current `main`).

PR #2567, opened 2026-04-30
fix: collapse caret to focus side of deleted range

christianhg:
After a `delete` event on an expanded selection, the caret now lands on the focus side of the original range rather than the anchor side. In all existing scenarios the visible cursor position is unchanged — both endpoints describe the same caret location at span boundaries — but the resulting `editor.selection` path matches the user's pointing direction. Downstream consumers that inspect the selection after a delete now see the side of the range the user was pointing at.

`applyDelete` grows a `backward` flag that travels alongside the range. `deleteRange` reads the original direction with `isBackwardRange` before `resolveExplicitRange` normalises the endpoints; `deleteCollapsed` derives it from the keystroke direction. The previously-named `'collapse-to-start'` selection mode is renamed to `'collapse-to-focus'` to match the new semantics.

The matrix test pins this through `boundaryEquivalentSelections` so it accepts either structurally-equivalent selection at a span boundary, and one explicit selection-shape assertion was updated to land on the focus side.

### DIFF
--- a/.changeset/delete-collapse-to-focus.md
+++ b/.changeset/delete-collapse-to-focus.md
@@ -1,0 +1,7 @@
+---
+"@portabletext/editor": patch
+---
+
+fix: collapse caret to focus side of deleted range
+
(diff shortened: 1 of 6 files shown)


christianhg:
Closing — research showed both ProseMirror and Lexical collapse to the start of the deleted range in document order, which is what PTE already does. The previously-named `'collapse-to-start'` semantics are the convention.

PR #2568, opened 2026-04-30
feat: container GA

christianhg:
This is the final wave of container work before GA. Lands the internal `defineLeaf` API, the rest of the engine container-awareness (behaviors, operations, converters, event resolution that hadn't caught up yet), and the playground schemas that exercise the full surface end-to-end. Salvaged from #2553.

After this, the container API holds together as a real feature: scope-based registration of editable fields, full DOM control for spans/inline-objects/void-block-objects, and behaviors that respect each container's sub-schema instead of falling back to the root.

## Engine

**`defineLeaf` and the public container API**

- Internal `defineLeaf` API + container-aware renderers. Consumers register `defineLeaf({scope, render})` for spans, inline objects, and void block objects to take full control over the outermost element rendered for matching nodes. Most specific JSONPath scope wins.
- Container wrappers emit `data-block-type="container"` instead of `data-slate-node="element"`, and the DOM inside a registered container is minimal: no inner text-block wrappers, no `pt-*` classes, no internal `data-child-*`. Engine queries that resolve DOM events to model paths accept the new attribute set.
- `defineLeaf` and `LeafPlugin` graduate to `@alpha` exports alongside `defineContainer`.
- `registerContainer` and `registerLeaf` move to the public `Editor` type as `@alpha`. The `InternalEditor` type is gone, so `useEditor()` returns `Editor` and the plugins no longer cast.

**Behaviors and operations finally container-aware**

- Decorator, annotation, list, and style lookups in behaviors scope to the container's sub-schema rather than the root. Toggling a decorator inside a callout consults the callout's declared decorators, not the root's.
- Converters, event-resolution, and block-offset utilities are container-aware. DOM events that land on a node inside a container resolve to a path that walks through the container's editable field.
- Cross-container decorator, style, and list-item toggles round-trip cleanly when the selection spans multiple sub-schemas: applied per-block, not as a single set across heterogeneous shapes.
- `list-item` toggle applies per-block when the selection crosses sub-schemas with different list-item declarations.
- Pressing Enter on a trailing empty line with an empty previous sibling escapes the editable container.
- Backspace doesn't consume the keystroke when the default style isn't available.

**Plugin extensibility**

- New `@beta` availability selectors (`isAvailableDecorator`, `isAvailableList`, `getAvailableDecorators`, `getAvailableLists`, `getAvailableStyles`) for external plugins that need to know what the current sub-schema declares.

## Playground

The playground schema now exercises every container shape we want to support at GA:

- Callout, table, code-block, and fact-box container examples covering list rendering, the `$..callout.block` second-container pattern, and per-leaf rendering inside containers.
- Container archetype stress test in the playground schema.
- Markdown-paste deserializer (paste markdown text and it lands as portable text).
- Fullscreen mode (CSS overlay modal, the editor never re-mounts).
- Markdown preview wired for the full playground schema.
- Fact-box surfaces in the toolbar and slash-command menu.

## Tests

22 new test files covering selectors, behaviors, operations, arrow navigation, break, split, decorator, mark state, move, range, select, annotation, and remote-value updates inside containers. Editor-state assertions use `toTextspec(snapshot.context)` for full deep-equality including caret position; selector tests assert path/key shape directly.

No public API breakage. `defineLeaf` and `defineContainer` remain `@alpha` through GA: the API is ready for real consumers to start writing against.


christianhg:
Superseded by the split:
- #2608 — defineLeaf + container DOM + playground archetypes
- #2609 — container-aware behaviors + availability selectors (depends on #2608)

PR #2571, opened 2026-04-30
fix: unset structural containers fully covered by a delete range

christianhg:
When a delete range covers an entire editable container — for example, the user clicks at the start of the first cell and shift-clicks at the start of the last cell of a 2x3 table — the cross-parent delete primitive trimmed the endpoints and applied the structural-container heuristic only to siblings between the start and end branches at the LCA. The walks up each branch still removed siblings as if their parent's field accepted text-blocks. Cells inside a row got dropped instead of having their contents cleared, and a fully-covered table was left as an empty shell.

Two changes in `deleteCrossParentRange`:

The walks up the start and end branches now check whether each level's parent field accepts text-blocks. When the parent is a structural container (a row whose `cells` field only accepts cells, a table whose `rows` field only accepts rows), trailing/preceding siblings get cleared via `clearContainerContents` instead of unset. The structural shell stays; only its contents become a fresh empty placeholder block.

A pre-check at the operation entry point walks up from the lowest common ancestor of the input range. For each ancestor that's a registered editable container, it compares the container's deepest start and end leaf points against the range's start and end. The outermost container that matches end-to-end gets unset wholesale. The check runs on the pre-unhang range so a cell-aligned selection still resolves to its original boundaries.

Adds three regression tests:
- A 2x3 empty table fully selected resolves to a single empty text block.
- The first two of three empty cells in a row is a no-op (nothing to delete).
- A selection from the first cell to a middle cell of a 2x3 table preserves every cell shell with its content cleared.

Updates two existing tests whose old expectations encoded the structural-trim bug (the "callout-in-cell-in-table" pair).

### DIFF
--- a/.changeset/unset-fully-covered-container.md
+++ b/.changeset/unset-fully-covered-container.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: unset structural containers fully covered by a delete range
--- a/packages/editor/src/internal-utils/delete-internal.ts
+++ b/packages/editor/src/internal-utils/delete-internal.ts
@@ -342,10 +342,17 @@ function deleteCrossParentRange(
   // 2. Walk up the start branch. At every level above the start block
   //    and below the start branch root, drop all trailing siblings.
   //    Trailing siblings of the start branch root itself are handled
-  //    by `removeChildrenBetween` below.
+  //    by `removeChildrenBetween` below. When the level's parent is a
+  //    structural container (its field doesn't accept text-block), the
+  //    siblings are structural shells (e.g. table cells in a row);
+  //    clear their contents instead of removing the shells themselves.
   let startLevel = startBlockPath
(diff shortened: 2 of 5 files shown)


christianhg:
@copilot Please rename the test on this PR's branch (`fix/delete-fully-covered-container`) at `packages/editor/tests/cross-container-range-delete.test.tsx`:

```
test('Selecting an entire empty table and pressing Delete unwraps the table', async () => {
```

→

```
test('Selecting an entire empty table and pressing Delete unsets the table', async () => {
```

This is the only change. The `fully-covered container` pre-check ends in `editor.apply({type: 'unset', path: ...})` — `unset` is the verb that matches the implementation; `unwrap` belongs to the empty-container cascade flow which moves children up.

Push as a fixup commit (`git commit --fixup=24e882f4d`) so it autosquashes cleanly when merged. No changeset needed (test-only rename).

Use Christian's git author/email: `Christian Hamburger Grøngaard <christian.groengaard@sanity.io>`.

PR #2581, opened 2026-05-01
fix: defer selection validation when the DOM hasn't caught up to the model

christianhg:
When Canvas opened sanity-io/canvas#1243 they hit the editor's `validate-selection-machine` catch path and worked around it on their side: pre-emptively clearing selection with `select({at: null})`, then re-selecting the new block in a `queueMicrotask` after their action set commits. The comment ([discussion](https://github.com/sanity-io/canvas/pull/1243#discussion_r3173223945)) calls it "a peculiar PTE behaviour where it will select the start of a document if it's unable to validate a selection." That's exactly what the editor was doing, and it's not the right shape.

## What was happening

`validate-selection-machine.ts` runs whenever the editable's `MutationObserver` fires. It compares the editor's model selection against what `toDOMRange` produces from the live DOM. When the model and DOM are in sync, the function syncs the DOM selection and returns. When `toDOMRange` throws, the catch path used to deselect and call `applySelect(start(editor, []))` — replacing whatever selection the consumer set with a forced cursor at the top of the document.

Action sets that synchronously mutate structure (insert a new block, then delete the original) trip this. The model has both operations applied; the DOM still shows the pre-action structure because React hasn't committed. `toDOMRange` walks the model selection's path, can't find the matching DOM nodes, throws. The catch fires and the consumer's intended selection is gone.

## What this changes

The catch now treats a `toDOMRange` throw as "DOM hasn't caught up yet" instead of "selection is invalid." `validateSelection` returns `'retry'`; the machine action re-fires the validation event in a `queueMicrotask`, giving React a tick to commit. On the next attempt the DOM matches the model, validation succeeds, and the selection survives intact.

If validation throws three times in a row the machine stops retrying. It leaves the selection alone — no forced deselect, no top-of-document fallback. If the selection is genuinely broken at that point, the next user interaction will trigger fresh DOM events and a fresh validation cycle.

## On testing

The catch only fires under real React render timing, where the model commits ahead of the DOM. In `vitest-browser` the operation queue, `MutationObserver`, and React commit interleave synchronously enough that the catch never fires for any scenario I could construct, including the exact action-set shape Canvas uses. The pinning test in `tests/validate-selection-action-set.test.tsx` verifies the action-set contract (selection lands on the new block) and passes on both pre-fix and post-fix code; it documents the desired behavior but doesn't discriminate the bug.

The static analysis is solid (the catch is unreachable in current tests, the retry path is straightforward), the existing `validate-selection-machine.test.ts` still passes, and the verification of the actual symptom can be done by removing Canvas's workaround and re-testing in their environment.

### DIFF
--- a/.changeset/validate-selection-defer.md
+++ b/.changeset/validate-selection-defer.md
@@ -1,0 +1,11 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: defer selection validation when the DOM hasn't caught up to the model
+
+When a behavior runs an action set that synchronously mutates structure (insert + delete a sibling, replace a block), the editor's `MutationObserver` fires while React is still mid-commit. The selection validation machine then runs against a model that's ahead of the DOM, `toDOMRange` throws, and the catch path used to deselect and re-select the top of the document, clobbering any selection the action set placed.
+
+The validation machine now treats the throw as "DOM hasn't caught up yet" rather than "selection is invalid." It re-fires itself in a microtask, giving React time to commit. After three failed retries it stops, leaving the selection alone instead of forcing it to the top of the document.
+
+Consumer behaviors that previously had to pre-emptively `select(null)` and `queueMicrotask(re-select)` to work around this can drop the workaround.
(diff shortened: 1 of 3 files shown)


christianhg:
Closing: the bug no longer reproduces. This PR's scenario (an action set that swaps the selected block keeps the selection on the new block) passes on current `main` once the expected value drops the since-evicted defaults — the destructive catch path in `validateSelection` (deselect + select top of document) was replaced during the selection-validation rework by a benign skip that lets the MutationObserver retry once React commits. Verified by running this PR's own test against `main` without the fix. The test lands separately as a regression pin.

PR #2582, opened 2026-05-01
feat!: remove editor._internal runtime field

christianhg:
Removes the `editor._internal` runtime field and drops the `editor` parameter from `useEditorSelector`.

`useEditorSelector(selector)` now resolves the editor from React context instead of taking the editor as a parameter. Every consumer audited (PTE itself, Canvas, plugin-typeahead-picker) pairs `useEditor()` immediately above `useEditorSelector(editor, ...)` — the parameter was ceremonial.

```diff
- const value = useEditorSelector(editor, selectors.getActiveStyle)
+ const value = useEditorSelector(selectors.getActiveStyle)
```

`createInternalEditor` returns a flat `{actors, editor, slateEditor, editable, subscriptions}` instead of nesting `slateEditor` and `editable` under `_internal`. Internal modules consume these fields directly through existing React contexts (`EditorActorContext`, Slate's `useSlateStatic()`).

The internal Slate editor and editable surface previously exposed via `editor._internal` are no longer reachable from outside the editor package. If you reached into `editor._internal` to drive raw machine events or read internal state, please open an issue with your use case.

### DIFF
--- a/.changeset/remove-editor-internal.md
+++ b/.changeset/remove-editor-internal.md
@@ -1,0 +1,9 @@
+---
+"@portabletext/editor": major
+---
+
+feat!: remove `editor._internal` runtime field
+
+The `_internal` field that exposed `editorActor`, `slateEditor`, and `editable` on the runtime editor object has been removed. Consumers that reached into this field through type casts (`editor as unknown as {_internal: ...}`) will get `undefined` at runtime in v7.
+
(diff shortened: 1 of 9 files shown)


christianhg:
Closing in favor of a cleaner approach. The WeakMap registry was indirection without justification — `editorActor`, `slateEditor`, `editable` are all in scope of `createInternalEditor`, and `useEditorSelector` already has `EditorActorContext` available. The right design is to thread these directly through return values / props / context instead of looking them up via a registry.

Will reopen as a fresh PR.

PR #2613, opened 2026-05-06
refactor: remove `ReadOnlyContext`

christianhg:
The `ReadOnlyContext` had a single consumer (`object-node.tsx`) and the editor's `readOnly` state already lives on the editor actor as the canonical source of truth.

The one consumer used it to omit `contentEditable={false}` on inline objects when the editor was readOnly — but the wrapping editable div is `contentEditable={!readOnly}`, so any child setting `contentEditable={false}` is a no-op when the parent is already non-editable. Always setting it on inline objects produces identical browser behaviour and lets us drop the context, the provider, and the hook entirely.

The diff is much larger than the meaningful change because removing the `<ReadOnlyContext.Provider>` wrapper re-indents about 800 lines of JSX. Append `?w=1` to the diff URL to ignore whitespace — the semantic change is `+2 / -10`.


christianhg:
Closing — the equivalence claim was wrong, the readOnly check on inline objects is load-bearing for `hasEditableTarget` and DOM-to-Slate point math via `closest('[contenteditable="false"]')` lookups. Keeping `ReadOnlyContext` as-is.

PR #2627, opened 2026-05-07
feat: accept the deprecated function-form on Editor.subscribe

christianhg:
`useSelector(editor, selector)` from `@xstate/react` works at runtime — xstate only calls observer-form `subscribe` and `getSnapshot`, both already on `Editor`. The blocker was purely at the type level: xstate's `Subscribable<T>` constraint also includes the deprecated function-form `subscribe(next, error?, complete?)`, which `Editor.subscribe` didn't expose.

This PR adds the function-form as a `@deprecated` overload on `Editor.subscribe`. The implementation re-routes to the observer-form internally. With the overload in place, `useSelector(editor, selector)` accepts the editor directly with no cast.

The internal `useEditorSelector` no longer needs to cast to `Pick<AnyActorRef, ...>`.

Two test files lock the contract:

- `editor.subscribable.test-d.ts` pins the type-level shape: `editor.subscribe(observer)` returns `{unsubscribe}`, `editor.getSnapshot()` returns `EditorSnapshot`, `useSyncExternalStore` accepts the editor via the adapter shape from the release post, a manual rxjs `Observable` wrapper around `editor.subscribe` compiles, and `useSelector(editor, selector)` compiles directly.
- `editor.subscribable.runtime.test.tsx` mounts a component using `useSelector` against the live editor and asserts the selector observes value-length changes after typing.

The trade-off is one extra overload in the public `Editor` type, marked `@deprecated`. The alternative is keeping `subscribe` observer-only and asking every consumer who reaches for `useSelector` to cast at the call site.

### DIFF
--- a/.changeset/editor-subscribe-function-form.md
+++ b/.changeset/editor-subscribe-function-form.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': minor
+---
+
+`Editor.subscribe` now accepts both the observer-form (`{next, error, complete}`) and the deprecated function-form (`subscribe(next, error?, complete?)`). The function-form satisfies xstate's wider `Subscribable<T>` constraint, so `useSelector(editor, selector)` from `@xstate/react` accepts the editor directly without a cast.
+
+Internally the function-form re-routes to the observer-form. New code should prefer the observer-form; the function-form exists for structural compatibility with libraries that still type their constraints to include it.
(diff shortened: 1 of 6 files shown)


christianhg:
Closing — going with a single-sentence mention in the v7 release post instead of widening `Editor.subscribe` with a deprecated function-form overload. The runtime works, the cast is small, and direct `useSelector(editor, ...)` isn't a usage we expect at scale.

PR #2636, opened 2026-05-09
fix: resolve container scopes recursively when parsing textspec into portable text

christianhg:
Surfaced while building gherkin scenarios for a future structured-lists plugin (#2635) — `fromTextspec` falls back to void block-objects when a container nests deeper than the schema resolver's bounded candidate set, the same bug class fixed in `toTextspec` by #2633.

For a self-referential schema where a `list-item.content.of` references the parent `list`, the resolver pre-emits depth-2 candidates (`list`, `list.list-item`, `list.list-item.list`). Runtime data nested at depth 3+ used `containers.has(scopedName)` and `containers.get(scopedName)?.field`, which both miss for keys outside that set. Routes through `lookupContainer` now, picking up the scope-pattern fallback that was already wired for runtime rendering.

The new round-trip test in `from-textspec.test.ts` parses a list nested three levels deep and asserts that serializing it back via `toTextspec` produces the original input. Verified by breaking: reverting the `lookupContainer` swap in `convertContainerToPTE` makes the test fail (the deepest list-item drops out of the round-trip).

### DIFF
--- a/.changeset/from-textspec-recursive-containers.md
+++ b/.changeset/from-textspec-recursive-containers.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: resolve container scopes recursively when parsing textspec into portable text
+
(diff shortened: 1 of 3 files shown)


christianhg:
Folded into #2635 as a `test:` commit (no changeset, test-utils only).

PR #2639, opened 2026-05-10
fix: keep `drag.dragover` cheap and depth-aware

christianhg:
`drag.dragover` fires at 50–100Hz during an internal drag. Today the drop-indicator computation goes through the behavior pipeline as a pair of behaviors registered via `createDropPositionBehaviorsConfig`. The `drag.dragover` behavior recomputes `getDragSelection`, `getSelectedBlocks`, and `isSelectingEntireBlocks` on every event, even though all three derive from `dragOrigin.selection` which is stable for the entire drag. The same-block check is `dragged._key === dropFocus._key` - a depth-1 comparison from the same bug class as #2638 (keys are sibling-unique, not tree-unique).

The behavior shape was the wrong layer for what the code actually does. Behaviors compose synthetic events that transform the editor model. This handler reads position from a DOM event and writes to a React state setter via an `effect` action. It's a UI affordance, not a transform.

This PR moves the drop-indicator computation into `useDropPosition` directly, alongside the React state it drives. The dragged-blocks list and entire-blocks check are computed once with `useMemo` keyed on `internalDrag`. Per dragover does one `getEnclosingBlock` for the cursor's block plus an `Array.some` over the small dragged-blocks list, then `setDropPosition`. Same-block discrimination uses `pathEquals` instead of key compare, which closes the depth-shared-key bug class.

`Editable` calls the returned `updateDropPosition` from its existing `handleDragOver` callback, alongside the existing `editor.send({type: 'behavior event', ...})` forwarding (so consumer plugins listening on `drag.dragover` keep working).

`createDropPositionBehaviorsConfig` and the two behaviors it returned are deleted. The `DropPosition` type moves to `use-drop-position.ts`.

7 existing dnd browser tests pass on chromium. 5 unit tests for `resolveElementDropPosition` (the per-element matcher from #2638) pass.

PR #2555 (the WeakMap memo on `dragOrigin`) is superseded by this - closing.


christianhg:
Closing — after the fixup that restored the behavior-level preventDefault, this PR ends up doing the same guard compute twice per dragover (once in the behavior, once in the hook) without a real win. The depth-shared-key bug (`_key === _key` comparison missing the depth disambiguation #2638 fixed in the render layer) is a one-line swap in the existing behavior, not a restructure. Will open that as a focused fix.

PR #2652, opened 2026-05-12
fix: inherit active decorators when `insert.blocks` merges a fragment span at a caret in a text block

christianhg:
When a consumer behavior catches `deserialize.data` for `text/plain` and raises `deserialization.success` directly with a bare text block, the engine should still inherit the caret's active decorators onto the pasted spans. Today it doesn't, and that surfaces as a real bug.

Pasting "Sting" into a bold span lands the text unbolded:

```
Before: B: foo [strong:ba|r] buz
After:  B: foo [strong:ba]Sting[strong:|r] buz   ← wrong, "Sting" should be bold
```

The engine has an abstract behavior on `deserialize.data` (`text/plain`) that raises `insert.span` with `decorators: getActiveDecorators(snapshot)`. That's the correct path for inheriting active marks. But it's just one path. Consumer behaviors that intercept `deserialize.data` (the playground's markdown deserializer, the existing `Consumer behavior on deserialize.data can produce rich blocks` scenario, any future deserializer) raise `deserialization.success` directly and skip it. The resulting `insert.blocks` flows into the fragment-merge path, which inserts spans as-is.

The fix is to inherit at the merge layer too. `insertFragmentChildren` now looks at the span at the caret position and copies its decorators onto any fragment span that has no marks of its own. Spans that arrive with their own marks are left alone (a consumer that pasted formatted content shouldn't have it stripped). Annotation refs (markDef keys) are not inherited; those continue to be threaded explicitly via `event.annotations` on the `insert.span` pathway.

The TDD scenario lives next to the existing "Pasted text/plain inherits formatting from surrounding text" test in `text-plain-paste.test.tsx`. It mounts a consumer behavior that mirrors the playground's markdown-deserializer (catches `text/plain`, raises `deserialization.success` with a bare text block), places the caret inside a bold span, pastes, and asserts the pasted text comes out bold. The test fails on main and passes with this change. Verified by breaking the fix (commenting out the decorator-filter) - the test reverts to red.

Reported via handy.computer pasting into the v7 playground - clipboard `text/plain` paste lost active formatting. Bug class predates v7 (latent since `insertFragmentChildren` was written) but only surfaces when a consumer plugin claims `deserialize.data` ahead of the engine's abstract behavior.

### DIFF
--- a/.changeset/insert-blocks-inherit-active-decorators.md
+++ b/.changeset/insert-blocks-inherit-active-decorators.md
@@ -1,0 +1,9 @@
+---
+"@portabletext/editor": patch
+---
+
+fix: inherit active decorators when `insert.blocks` merges a fragment span at a caret in a text block
+
+When the abstract `text/plain` deserializer raises `insert.span`, the pasted text inherits the caret's active decorators. Consumer behaviors that bypass that path - for example a markdown deserializer that catches `deserialize.data` and raises `deserialization.success` directly with a block whose spans carry empty marks - went through the engine's fragment-merge path, which dropped the caret context and inserted the spans bare. Pasting inside a bold span would land the new text unbolded.
+
+The fragment-merge path now copies the decorators of the span at the caret onto any fragment span with no marks, matching the abstract `insert.span` behavior. Spans that arrive with their own marks are left alone. Annotation refs are not inherited; those continue to be threaded explicitly via `event.annotations`.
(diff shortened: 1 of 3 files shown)


christianhg:
Closing after an evidence-based revival attempt. The bug is alive: this PR's own test is red on current `main` (pasting `new` at a caret in `[strong:ba|r]` yields `[strong:ba]new[strong:|r]`). But the fix does not survive the operation refactors since March: transplanted onto today's `insertFragmentChildren`, the paste inserts nothing at all, so the inheritance logic needs re-deriving against the current internals, and the revised patch pipeline work is about to reshape this seam anyway. Tracked in Linear with the analysis. The branch preserves the pinning test and the design note that only decorators, not annotations, are inherited.

PR #2654, opened 2026-05-12
feat: fast-reject container lookups when the type is not a registered container

christianhg:
Container lookups during tree descent (`isEditableContainer`, `getNodeChildren`, `getEnclosingContainer`, normalization, `getDirtyPaths`, …) ask the same question for every node visited: "is the type at this scoped name a registered container?" For a deep nested tree, the answer is "no" for the overwhelming majority of nodes - text blocks, spans, leaves, inline objects - yet every one of them runs the full scope-walk fallback through `matchScope` and `compareSpecificity` to confirm it.

A static audit of a 30-iteration deep-nested-list harness counted `matchScope` at 309K calls and `compareSpecificity` at 208K calls. ~97% of those came through `lookupContainer` calls from `getNodeChildren`. The same audit showed that an exact-match fast path on `containers` already absorbs the common-case hit; what remained was a tax paid by every type that isn't a container in the first place.

This PR adds a `containerTypes: ReadonlySet<string>` sibling to `containers` on `EditorContext`. It holds the set of types that appear as the leaf segment of any registered container scope (e.g. `$..table.row.cell` contributes `cell`). `lookupContainer` consults the set before falling back to scope walking:

```ts
export function lookupContainer(
  containers: ResolvedContainers,
  containerTypes: ReadonlySet<string>,
  scopedName: string,
): ContainerConfig | undefined {
  const exact = containers.get(scopedName)
  if (exact) return exact
  const leaf = scopedName.includes('.')
    ? scopedName.slice(scopedName.lastIndexOf('.') + 1)
    : scopedName
  if (!containerTypes.has(leaf)) return undefined  // <-- new
  // ...scope-walk fallback
}
```

The set is derived in `resolveContainerTypes(configs)` whenever containers are (re)resolved - same three call sites in `editor-machine` that already rebuild `containers` (`register container`, `unregister container`, `sync containers`). No caching, no symbol-keyed slots, no WeakMaps - the set is constructed in O(scopes) and stored alongside `containers` on the context.

### What this is and isn't

This is a **negative-set fast-reject**: "is matching even possible at this leaf type?" It collapses the `lookupContainer` overhead for non-container nodes from O(scopes) to O(1). It does not change the answer the function returns; the scope-walk fallback runs for every type that *could* match.

This is not the held-back terminal-keyed positive-index redesign - that one tries to answer "which scope matches" without backtracking. They are different optimizations with different evidence. We're shipping this one because the audit data justifies it directly.

### Production impact

The 30-iteration harness counts `matchScope` collapse from 309K → ~1K, and `compareSpecificity` from 208K → ~1K. Whether that translates to a visible wall-clock win in production depends on what fraction of the previously-reported `matchScope` self-time was real per-call cost versus CPU-sampler attribution. Either way, the function drops out of profiles.

### API surface

`containerTypes` is exposed as `@alpha` on `EditorContext` alongside `containers`. Consumers reading the snapshot can rely on it being present when `containers` is. The naming follows the user-perspective ("is this a container type") rather than the parser jargon ("terminal segment"); thanks to @christiangroengaard for the prompt.

### DIFF
--- a/.changeset/perf-lookup-container-fast-reject.md
+++ b/.changeset/perf-lookup-container-fast-reject.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': minor
+---
+
+feat: fast-reject container lookups when the type is not a registered container
+
+`lookupContainer` now consults a new `containerTypes: ReadonlySet<string>` sibling on `EditorContext` and returns `undefined` without scope-walking when the candidate type isn't registered as a container anywhere. Most calls during a tree descent hit non-container types (spans, text blocks, leaves), so the new set short-circuits them in O(1) instead of falling through to scope matching.
+
+`containerTypes` is exposed as `@alpha` alongside `containers`. The set is rebuilt from the parsed scopes whenever containers are (re)resolved.
--- a/packages/editor/src/editor/create-slate-editor.tsx
+++ b/packages/editor/src/editor/create-slate-editor.tsx
@@ -25,6 +25,7 @@ export function createSlateEditor(
     context: {
       schema: context.schema,
       containers: context.containers,
+      containerTypes: context.containerTypes,
       value: [],
       keyGenerator: context.keyGenerator,
     },
@@ -36,6 +37,7 @@ export function createSlateEditor(
   editor.schema = context.schema
   editor.keyGenerator = context.keyGenerator
   editor.containers = new Map()
+  editor.containerTypes = new Set()
   editor.leafs = new Map()
 
   editor.decoratedRanges = []
--- a/packages/editor/src/editor/editor-machine.ts
+++ b/packages/editor/src/editor/editor-machine.ts
@@ -28,6 +28,7 @@ import type {
 import {
   resolveContainerField,
   resolveContainers,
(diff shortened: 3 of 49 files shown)


christianhg:
Superseded by the container API redo at `/specs/container-api-redo.md` (approved by Christian as p0).

The redo deletes the scope grammar entirely, replacing scope-walk lookup with `Map<_type, Container | Leaf>` direct lookup. The `containerTypes` fast-reject this PR introduced optimizes a lookup that no longer exists, so this PR closes.

Follow the redo on `feat/container-api-redo`.

christianhg:
Superseded by #2655. The container API redo deletes scope grammar entirely (`lookupContainer`, `matchScope`, `compareSpecificity`, the `scope/` directory). The `containerTypes` fast-reject this PR added accelerates a lookup that no longer exists, so it's not a partial step toward the redo - it's parallel work that the redo subsumes by deletion.

The redo's first-principles answer to "is matching possible?" is "containers and leaves are keyed by `_type` directly; lookup is two `Map.get` calls." No fast-reject needed.

PR #2655, opened 2026-05-13
feat: `renderChild` on `defineContainer` and a type-keyed container/leaf registry

christianhg:
## Motivation

The container API has been in development for a while. The earlier shape - the one this PR replaces - addressed types positionally in the schema tree via a JSONPath-flavored scope grammar:

```ts
const calloutContainer = defineContainer({
  scope: '$..callout',
  field: 'content',
  render: CalloutRender,
})

const calloutImageLeaf = defineLeaf({
  scope: '$..callout.image',
  render: CompactImage,
})

const imageLeaf = defineLeaf({
  scope: '$..image',
  render: FullImage,
})

const tableCellImageLeaf = defineLeaf({
  scope: '$..table.row.cell.image',
  render: CompactImage,
})
```

Each scope was parsed, walked against the runtime type chain, ranked by specificity, and resolved on every render. The grammar conflated five capabilities in one mechanism: positional addressing, cross-plugin override, multi-level depth, wildcard descent, and specificity ranking. Spending time with it in the playground surfaced three problems:

1. **Cross-plugin override has no analog in the wider ecosystem.** TipTap, Lexical, Plate, ProseMirror, and Slate all let plugins register renderers for ONE type globally; none offer "plugin A registers a render that fires inside plugin B's container." The capability would be unique to this editor for non-trivial reasons (override wars, ordering bugs, mental-model overhead). The playground had 15 multi-segment scope registrations; exactly 1 used cross-plugin reach. The other 14 expressed the same pattern: "this plugin owns its container AND what things look like inside it" - which the grammar enabled but didn't enforce.

2. **Scope resolution is on the hot path.** `matchScope` + `compareSpecificity` ran on every element render. Deep traces flagged scope resolution as a significant share of in-keydown self-time. The redesign closes the cost path entirely either way: type-keyed lookup is two `Map.get` calls.

3. **Schema position grammar is in the wrong place.** Where a type can be used is a schema concern, not a registration concern. The scope grammar duplicated that information on the plugin side, which let plugins register types at positions the schema didn't permit (a no-op) and forced authors to remember a second grammar layered on top of the one they already wrote.

Net diff is removal-biased: -4,717 lines deleted, +2,076 added. The grammar, walk, and specificity machinery is gone. The container API stays at `@alpha` - nothing externally released changes shape, this is the next iteration to try.

## Design principle

A plugin owns its container AND what things look like inside its container.

Registration is keyed by `_type`. Plugins declare what their type IS (via the schema) and how it should RENDER (via `defineContainer` / `defineLeaf`). Positional control over what's inside a container lives on that container's own registration via `renderChild` - a per-child-type render record. The scope grammar deletes; the schema is the single source of truth for type placement.

This trades one capability for a simpler model: cross-plugin positional reach is gone. The one playground case that used it (image plugin's compact-images-in-table-cells) moves into the table plugin's cell registration where the domain ownership is structural rather than policy.

### Register a container

```ts
// before
const calloutContainer = defineContainer({
  scope: '$..callout',
  field: 'content',
  render: CalloutRender,
})

// after
const calloutContainer = defineContainer({
  type: 'callout',
  childField: 'content',
  render: CalloutRender,
})
```

The result is passed to `ContainerPlugin` exactly as before - via the `containers` prop, alongside any other container registrations the plugin ships.

### Style a child type inside a container

```ts
// before - the callout plugin reaches into the global image leaf via
// a callout-scoped scope string, on top of the image plugin's own
// global registration.
const calloutContainer = defineContainer({
  scope: '$..callout',
  field: 'content',
  render: CalloutRender,
})
const calloutImageLeaf = defineLeaf({
  scope: '$..callout.image',
  render: CompactImage,
})
const imageLeaf = defineLeaf({
  scope: '$..image',
  render: FullImage,
})

// after - callout plugin declares its container AND its per-child renders
// together. No cross-plugin reach.
const calloutContainer = defineContainer({
  type: 'callout',
  childField: 'content',
  render: CalloutRender,
  renderChild: {image: CompactImage},
})
const imageLeaf = defineLeaf({
  type: 'image',
  render: FullImage,
})
```

### Cross-plugin reach (gone after this PR)

```ts
// before - image plugin reached into table's domain
const imageLeaf = defineLeaf({scope: '$..image', render: FullImage})
const tableCellImageLeaf = defineLeaf({
  scope: '$..table.row.cell.image',
  render: CompactImage,
})

// after - table plugin owns what's inside its cells
const cellContainer = defineContainer({
  type: 'cell',
  childField: 'content',
  render: CellRender,
  renderChild: {image: CompactImage},
})
const imageLeaf = defineLeaf({type: 'image', render: FullImage})
```

### A child type the parent plugin alone knows about

`renderChild` fires whether or not the child has a global registration. A `chart` type that only appears inside callouts can be rendered entirely from the callout plugin - no separate `defineLeaf` registration.

```ts
const calloutContainer = defineContainer({
  type: 'callout',
  childField: 'content',
  render: CalloutRender,
  renderChild: {chart: ChartView},
})
```

### Override text-block rendering globally

`block` is a type like any other. Register it as a container and the override replaces the default text-block pipeline (style switching, list-item rendering, decorator/annotation wiring) for every text block in the document.

```ts
const blockContainer = defineContainer({
  type: 'block',
  childField: 'children',
  render: MyBlockRender,
})
```

### Override text-block rendering inside a specific container

Symmetric with the global case: `renderChild.block` replaces the default pipeline for blocks that are direct children of this container only.

```ts
const codeBlockContainer = defineContainer({
  type: 'code-block',
  childField: 'lines',
  render: CodeBlockRender,
  renderChild: {block: CodeLine},
})
```

Same dispatch path, same author responsibility. Both fully replace the default pipeline; the only difference is scope.

## Engine dispatch

For each rendered element, the engine consults configs in this order:

```
parent.renderChild?.[child._type]   // positional override on the immediate parent
  ?? containers.get(child._type)    // global container registration
  ?? leaves.get(child._type)        // global leaf registration
  ?? engineDefault                  // built-in pipeline (text blocks, void objects)
```

`renderChild` keys off the IMMEDIATE structural parent, not an ancestor walk. An image inline inside a text block inside a callout goes through the global image registration, not the callout's `renderChild.image` - the callout is a grand-ancestor at that point, not the direct parent.

`renderChild` fires regardless of whether the child has a global registration. The parent plugin is self-contained: removing it returns the child to whatever else registered the type (or engine default).

`renderChild.block` and global `defineContainer({type: 'block'})` are both full-replace. Authors who want partial override use CSS scoped from the wrapper, or the consumer-level `renderBlock` / `renderStyle` / `renderListItem` props on the editable.

## Multi-level containers (tables)

Each container declares ONE editable array via `childField`. Multi-level nesting works by chaining registrations: each level's render owns its own dispatch boundary; the engine walks the declared chain.

```ts
const tableContainer = defineContainer({
  type: 'table',
  childField: 'rows',
  render: TableRender,
})

const rowContainer = defineContainer({
  type: 'row',
  childField: 'cells',
  render: RowRender,
})

const cellContainer = defineContainer({
  type: 'cell',
  childField: 'content',
  render: CellRender,
  renderChild: {image: CompactImage},
})
```

For a document containing a `table` with rows of cells, the engine dispatches level-by-level:

```
root child is _type: 'table'
  -> containers.get('table') -> TableRender fires
  -> TableRender's children prop is node.rows (because childField: 'rows')
  -> for each row, dispatch fires again

row -> containers.get('row') -> RowRender
  -> children prop is node.cells

cell -> containers.get('cell') -> CellRender
  -> children prop is node.content
  -> for each content item:
      cell.renderChild?.[item._type]  // CompactImage if item._type === 'image'
        ?? containers.get(item._type)
        ?? leaves.get(item._type)
        ?? engineDefault
```

No tree-walk to "find" cells. No wildcard matching. Just local dispatch at every parent-child boundary, with the immediate parent's `renderChild` as first priority.

## Scope of PTE dispatch

**PTE only sees objects inside arrays that some container declared as its `childField`.** The PTE-managed tree is exactly the transitive closure of `childField` walks from the root.

Non-`childField` data on a node is opaque to the engine. The container's render function receives the full `node` and is free to render other fields - a `caption` string, a `metadata` object, an unrelated `shapes` array - using the consumer's own React. The engine does not dispatch into those arrays, doesn't walk them, doesn't apply container or leaf registrations to objects inside them.

A consequence: registering `cell` as a container makes it a PTE container wherever it appears as a child of an array some container declared as its `childField` (and the schema permits it). It does NOT make `cell` render anywhere a cell-shaped object happens to exist on some other property. If `shapes` was declared as a `childField` somewhere, cells inside it ARE PTE containers and the same `CellRender` fires; if not, those cells are opaque data the consumer renders themselves.

Scope grammar's `$..cell` wildcard descent had no such boundary - under the old shape, PTE would have matched cells anywhere they appeared in the value tree. Now the boundary is structural: the `childField` chain defines exactly what PTE recurses into.

## Same type in two structural roles

A real example of why `renderChild` exists as an API. Suppose the schema permits `cell` in two places: `row.cells[]` (table cells) and `diagram.shapes[]` (cells used as shapes in a diagram). The consumer wants them rendered differently. Three patterns work:

**Rename in the schema.** If you control the schema and renaming clarifies intent, call them `cell` and `shape-cell`. Two separate global registrations, type system disambiguates, no engine logic at all. The cleanest answer when it's available.

**`renderChild` on each parent.** When renaming isn't an option (legacy schema, third-party data, or `cell` legitimately is the same concept in two roles):

```ts
const rowContainer = defineContainer({
  type: 'row',
  childField: 'cells',
  render: RowRender,
  renderChild: {cell: TableCellRender},
})

const diagramContainer = defineContainer({
  type: 'diagram',
  childField: 'shapes',
  render: DiagramRender,
  renderChild: {cell: ShapeCellRender},
})
```

No global `defineContainer({type: 'cell'})`. Each parent intercepts via its own `renderChild.cell`. Engine never reaches a global fallback because the parent's `renderChild` is consulted first. Table cells get `TableCellRender`, diagram cells get `ShapeCellRender`, neither parent knows about the other.

**Branch inside a single global render on `parent?._type`.** Works, but mingles positional behavior with type-identity in the render function. Available, not recommended.

The redesign supports the first two cleanly and leaves the third as an escape hatch. The trade-off compared to scope grammar's specificity-based ranking: positional differentiation is now a parent's responsibility (declared in `renderChild`), not a free-floating registration anyone can attach.

## Compared to scope grammar

| | scope grammar (before) | type-keyed + renderChild (after) |
|---|---|---|
| Levels | multi-level (`$..code-block.block.span`) + wildcard descent | direct parent only |
| Where it lives | anywhere - any plugin can register any scope | only on the container's `defineContainer` |
| Cross-plugin reach | yes - image plugin can register `$..callout.image` | no - only callout plugin can declare what's inside callout |
| Specificity | `compareSpecificity` ranks multi-matches | impossible by construction - one match max per (parent, child) |
| Hot path | parse + match + walk + compareSpecificity per render | two `Map.get` calls |
| Type narrowing | none | `defineContainer<typeof schema>({...})` narrows `type`, `childField`, `renderChild` keys |
| Same type, different render by position | yes, via separate scopes | `renderChild` on each parent, or different type names |
| Registering same type twice | both win positionally per specificity | first wins; second logs a warning |

### `defineContainer`

```ts
defineContainer({
  type: string                              // required, the _type to register
  childField: string                        // required, schema field holding editable children
  render?: ContainerRender                  // optional - engine provides a default wrapper
  renderChild?: {[childType]: ChildRender}  // optional, per-child-type overrides
})
```

`render` and `renderChild` entries receive `{attributes, children, focused, node, parent, path, readOnly, selected}`. `parent` is the immediate parent node reference, referentially stable across re-renders, undefined at root. Container renders never receive `isInline` - containers always sit in block position.

With a schema type parameter, `type` narrows to registered container type names, `childField` narrows to that type's array field names, and `renderChild` keys narrow to child types valid in that field per the schema:

```ts
const calloutContainer = defineContainer<typeof schema>({
  type: 'callout',
  childField: 'content',
  renderChild: {image: CompactImage},  // <- keys checked against schema
})
```

### `defineLeaf`

```ts
defineLeaf({
  type: string                              // required
  render: LeafRender                        // required
})
```

`render` receives `{attributes, children, focused, isInline, node, parent, path, readOnly, selected}`. Leaves can be inline (sitting inside a text block's `children`) or block (sitting in an array of blocks); `isInline` discriminates. The same leaf type can be registered as both - the render branches.

### Composition

Unchanged. `ContainerPlugin` accepts a `containers` prop, `LeafPlugin` accepts a `leaves` prop, each taking an array of `defineContainer` / `defineLeaf` results. A plugin component returns one or both:

```ts
function TablePlugin() {
  return ContainerPlugin({
    containers: [tableContainer, rowContainer, cellContainer],
  })
}
```

(Real plugins return JSX, omitted here to keep code blocks portable.)

## Playground rewrite

The playground was the corpus this PR was sized against. 14 of its 15 multi-segment scope registrations rewrite mechanically using the patterns above. Three were already redundant and delete:

- `$..callout.block` - style switching that the default text-block pipeline already does, scoped via the outer wrapper
- `$..fact-box.block` - same shape
- `$..code-block.block.span` - `pre` inheritance handles monospace

One test (`scoped block overrides universal block inside container`) deletes alongside. It registered two `defineContainer({type: 'block'})` and relied on specificity to resolve the conflict. With the new shape, first wins and the second registration logs a warning. Documented semantic shift, not a regression.

## Trade-offs

**Lost capability: cross-plugin positional override.** A plugin can no longer register a render that fires inside another plugin's container without coordination. The playground had one such case (image-in-table-cell); it moves cleanly to the table plugin owning the override. If a future case can't be expressed in `renderChild`, the available patterns are:

- Type owner branches on `parent?._type` inside its render
- Outer plugin wraps the inner render via `renderChild`
- React Context provided by the outer container, consumed by the inner render
- Render-prop on the inner plugin

**Lost capability: same type, different render by position via specificity.** The earlier shape let `$..image` and `$..callout.image` resolve differently by specificity. The new shape requires `renderChild` (positional inside one container) OR different type names (`compact-image` vs `image`).

**Schema position grammar is gone.** A registered container type applies wherever its bare `_type` appears in the value. If the schema permits `row` at the top level AND inside a table, both positions are containers. Authors who want position-differentiated structure use different type names.

Supersedes #2654.


christianhg:
Closing this — superseded by the Container API v2 redesign. See `/specs/container-api-v2.md` for the new shape and rationale; a fresh PR will follow.

The short version: this PR's `renderChild` map kept registrations on `defineContainer` as a positional override, but conversations with @snorrees surfaced that **containers and leaves recursively all the way down** is the cleaner shape. The v2 proposal nests `defineContainer` / `defineLeaf` in a parent's `of` array, same primitive at every level, no second-class registration surface.

Thanks for the review on this iteration — the dual-`cell` problem you flagged is what unblocked the redesign.

PR #2686, opened 2026-05-20
fix: gate `createHeadingRule` on the resolved style being present in the sub-schema

christianhg:
The heading input rule for the markdown shortcuts plugin fires whenever `/^#+ /` matches at the start of the focus block, including in sub-schemas that do not declare any of the styles the consumer's `headingStyle` callback returns. The rule already bails when the callback returns `undefined`, but a buggy callback that returns a style name unconditionally - rather than guarding on whether it is in `context.schema.styles` - would cause the rule's actions to dispatch a `block.set` for an unknown style, swallowing the keystroke.

The rule now self-gates: after resolving the style via the callback, it checks that the style exists in the sub-schema at the focus. If not, the guard returns `false` and the input rule does nothing.

A new scenario in `sub-schema.test.tsx` pins the contract by using a deliberately buggy `headingStyle` callback that returns `'h1'` unconditionally inside a code-block sub-schema whose styles are `[{name: 'code'}]`. Typing `# ` now leaves the text as `"# "` instead of being consumed.

### DIFF
--- a/.changeset/heading-rule-codeblock-gate.md
+++ b/.changeset/heading-rule-codeblock-gate.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/plugin-markdown-shortcuts': patch
+---
+
+fix: gate `createHeadingRule` on the resolved style being present in the sub-schema
--- a/packages/plugin-markdown-shortcuts/src/rule.heading.ts
+++ b/packages/plugin-markdown-shortcuts/src/rule.heading.ts
@@ -52,6 +52,10 @@ export function createHeadingRule(config: {
         return false
       }
 
+      if (!subSchema.styles.some((s) => s.name === style)) {
+        return false
+      }
+
       return {match, style}
     },
     actions: [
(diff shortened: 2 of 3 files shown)


christianhg:
Closing — the framing is wrong.

`createHeadingRule` receives a consumer-supplied `headingStyle` callback that ALREADY gets `context.schema` (which IS the sub-schema at the focus block's path). The consumer can already gate by returning `undefined` when the style isn't in scope — pilcrow does exactly this:

```ts
headingStyle={({context, props}) => {
  const style = `h${props.level}`
  return context.schema.styles.some((s) => s.name === style)
    ? style
    : undefined
}}
```

Adding a second gate inside the rule that overrides what the consumer returned violates the consumer's authority. If they want `h1` to land at a path where `h1` isn't in the sub-schema, that's THEIR call (maybe they have an out-of-band style they're registering, maybe they want the schema to relax). The plugin shouldn't second-guess the explicit answer.

What I should've done in the original `c780f78` pilcrow fix was just document the gate as the consumer-side responsibility, not push it upstream. Apologies for the noise.

Closing.

PR #2687, opened 2026-05-21
fix: drag containers and block-objects via their decoration chrome

christianhg:
Dragging a container (callout, blockquote, code-block, table) or block-object (image, hr) by a `contentEditable=false` decoration inside its render output - such as a block handle or drag affordance - now moves the entire block.

Previously, the engine resolved the drag to the deepest text point inside the container, dropped the inner block at the destination, and left the container intact at the origin. The symptom: drag a callout, get the callout's first paragraph; the callout stays put.

### The bug had three independent layers

Each layer was individually sufficient. Fixing only one left the user-visible behavior unchanged.

**1. `getEventPosition` resolved chrome-dragstart to inner caret.**

When `dragstart.target` lands on a `contentEditable=false` decoration inside a block (the block-handle button rendered by the consumer), `getDomNodePath` walks up to the block's `data-pt-path` correctly. But `getSelectionFromEvent` then runs `caretPositionFromPoint(clientX, clientY)`, which ignores `contentEditable=false` and recovers a caret at the inner text directly under the cursor. That deeper selection overrode the block-level position.

Fix: a new `isEventOnChrome` walker detects whether the event target sits inside a `contentEditable=false` subtree under a `data-pt-path` ancestor. When it does, we keep the block-level resolution and skip the inner-caret recovery.

**2. `behavior.core.dnd` move-mimic used `getFragment`, which intentionally unwraps containers.**

`getFragment` returns "the smallest top-level-valid fragment of the editor's value that covers the current selection" - correct for clipboard copy, wrong for in-editor move. For a chrome-drag of a callout whose contents are themselves root-acceptable, it returned the inner paragraph, not the callout.

Switching to `getSelectedBlocks` (which preserves containers whole) blanket-style broke deeper drags: when the selection is inside a table cell, `getSelectedBlocks` returns the whole table (the root-level block covering the selection), not the inner block the user grabbed.

Fix: discriminate on origin path length. Top-level chrome drags (path length 1) use `getSelectedBlocks`; deeper drags - block-objects inside cells, multi-block selections inside containers - keep using `getFragment`. Both shipped tests for in-cell drag remain green.

**3. `operation.delete` with `unit:'block'` walked through `resolveSelection` and `unsetMatchedNodesInRange` dedup, which picked the descendant.**

`delete.block at: [{_key:'cl-24'}]` raises `delete` with a collapsed selection at the callout's path. The operation ran `resolveSelection`, which walks any non-text-block path *down to the deepest first text point*. That made the from/to range cover both `[cl-24]` (in range as ancestor of the resolved point) and `[cl-24, content, b-30]` (also in range). `unsetMatchedNodesInRange` then deduped by keeping the deepest candidate - which is correct for double-unset prevention on overlapping text-range deletes, but wrong here. The inner block got unset; the callout stayed.

Fix: short-circuit `operation.delete` when `unit:'block'` and `at` is collapsed on a block's path. Unset the explicit node directly, before resolveSelection runs.

### Tests

- New `event.drag.dragstart-from-chrome.test.tsx` mounts a callout with a `contentEditable=false` draggable button inside its render and dispatches a real `dragstart` at the button's coordinates. Asserts the position selection resolves to the callout's path, not a deep inner caret.
- All existing dnd tests pass (`event.drag.drop.test.tsx`, `event.drag.drop.self-drop.test.tsx`). Two scenarios that exercise deep table-cell drags - block-object between cells, multi-block selection inside cell - drove the path-length-1 guard on the move-mimic switch.

### Consumer-visible

Block handles rendered inside a block's render callback as `contentEditable=false draggable={!readOnly}` now drag the whole block, including containers. The pattern (handle as part of the block's render output, inside the Slate DOM tree) was already canonical for block-objects; this PR makes it work end-to-end for containers as well.

### DIFF
--- a/.changeset/fix-drag-from-chrome.md
+++ b/.changeset/fix-drag-from-chrome.md
@@ -1,0 +1,33 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: drag containers and block-objects via their decoration chrome
+
+Dragging a container (callout, blockquote, code-block, table) or
+block-object (image, hr) by a `contentEditable=false` decoration
+inside its render output - such as a block handle or drag affordance -
+now moves the entire block. Previously, the engine resolved the drag
+to the deepest text point inside the container, dropped the inner
+block at the destination, and left the container intact at the origin.
+The bug surfaced as "drag callout, get the callout's first paragraph;
+the callout stays put."
+
+Three independent layers needed fixing:
+
+- `getEventPosition` now recognizes when `dragstart.target` lives
+  inside a `contentEditable=false` subtree under a `data-pt-path`
+  ancestor (the decoration chrome) and resolves the position to the
+  block's own path instead of letting `caretPositionFromPoint`
+  recover a caret inside the block's text.
+- The internal-drag move-mimic uses `getSelectedBlocks` when the
+  drag origin is at the top level (path length 1) - which preserves
+  containers whole - and keeps using `getFragment` for deeper
+  origins (block-objects or text inside container cells), where
+  unwrapping to root-acceptable shape is still correct.
+- `operation.delete` with `unit:'block'` short-circuits when
+  `at` is collapsed on a block's path. The old flow ran
+  `resolveSelection` (which walks to the deepest first text point),
+  then `unsetMatchedNodesInRange` would yield both the explicit block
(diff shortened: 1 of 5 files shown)


christianhg:
Closing — paving over in pilcrow instead.

PR #2762, opened 2026-06-08
feat: extend `blockIndexMap` to every keyed node

christianhg:
`EditorSnapshot.blockIndexMap` now indexes every keyed node in the tree — root blocks, spans, inline objects, and nodes nested inside registered containers. Entries are keyed by the serialized full path from the document root.

Sibling-navigation selectors (`getPreviousBlock`, `getNextBlock`, `getNextSpan`, `getPreviousSpan`, `getNextInlineObject`, `getPreviousInlineObject`) and behavior `getSibling` calls now resolve in constant time at any depth. Before, only root blocks were indexed, and any sibling lookup deeper in the tree fell back to a `findIndex` scan over the parent's children.

## API

The public `EditorSnapshot.blockIndexMap` type stays `ReadonlyMap<string, number>`. A `BlockIndexMap` subclass overrides `get` and `has` to translate bare `_key` lookups into serialized paths, so external consumers that currently read by `_key` keep working without changes:

```ts
snapshot.blockIndexMap.get('k0')                // still works for root blocks
snapshot.blockIndexMap.get('[_key=="k0"]')      // also works
snapshot.blockIndexMap.get('[_key=="k0"].children[_key=="s0"]')  // new, any depth
```

Internal engine readers (`getNode`, `getNodes`, `getAncestors`, `getDomNode`, `modify`, child- and block-index resolution in `apply-operation`, and the uniqueness check in abstract insert) all read by serialized path.

## Maintenance

Map maintenance is consolidated into a single pure function `transformBlockIndexMap(map, op, beforeValue, afterValue, ctx)` called once per operation from the update-value plugin. It dispatches on op shape and only touches the affected subtree, so per-op work is bounded by the subtree size and not by map size.

### DIFF
--- a/.changeset/block-index-map-path-keyed.md
+++ b/.changeset/block-index-map-path-keyed.md
@@ -1,0 +1,15 @@
+---
+'@portabletext/editor': minor
+---
+
+feat: extend `blockIndexMap` to every keyed node
+
+`blockIndexMap` now indexes every keyed node in the tree (root
+blocks, spans, inline objects, and nodes nested inside registered
+containers). Entries are keyed by the serialized full path from the
+document root.
+
+Sibling-navigation selectors (`getPreviousBlock`, `getNextBlock`,
+`getNextSpan`, `getPreviousSpan`, `getNextInlineObject`,
+`getPreviousInlineObject`) and behavior `getSibling` calls resolve
+in constant time at any depth.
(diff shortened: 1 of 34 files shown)


christianhg:
Superseded by #2769, which shipped this in `5ee8bffd8` and published in `@portabletext/editor@7.5.0`. Closing this earlier draft as a duplicate.

PR #2764, opened 2026-06-09
feat: emit behavior events from `editor.on`

christianhg:
Today `editor.on` exposes a fixed set of events (`patch`, `mutation`, `selection`, plus lifecycle). The behavior events that flow through the chain — `insert.text`, `drag.dragover`, `serialize`, `select`, etc. — have no public subscription surface. A plugin author who wants to derive state from "the user is dragging over this position" can't do it cleanly: either they reach into engine internals, or they request a new `RenderProps` field.

This PR widens `EditorEmittedEvent` to include `BehaviorEvent`. Behavior events are emitted from the editor machine and forwarded through the relay, so consumers subscribe by name:

```ts
editor.on('drag.dragover', (event) => {
  // event.position, event.dragOrigin
})

editor.on('insert.text', (event) => {
  // event.text
})

editor.on('*', (event) => {
  // sees behavior events alongside the existing emitted events
})
```

`editor.on(type, listener)` is typed-narrowed via xstate's existing channel overload, so call sites get the right event shape without `event.type === '...'` guards. `EventListenerPlugin` sees behavior events too.

## Pre- vs post-handler tap

The tap fires once per behavior event, **before the behavior chain runs**. Observers see user intent regardless of whether a high-priority handler decides to swallow the event.

Post-handler would miss intercepted events, which is the wrong default for plugin authors building providers. Example: a `dropPosition` provider needs to track the latest `drag.dragover` even when `behavior.core.dnd` ends up swallowing the dragover to set its own internal state — the provider's view of "where the user is dragging" shouldn't depend on which handler wins.

A consequence of this placement: `raise(event)` calls inside an action body do *not* re-enter the `handle behavior event` action, so they're invisible to `editor.on`. Only externally-dispatched and `sendBack`-routed behavior events reach subscribers. That preserves the "observe intent" semantic (subscribers see the events users sent, not decomposition machinery).

## Removed: deprecated `'unset'` emitted event

The deprecated `'unset'` emitted event collided with the `BehaviorEvent` variant of the same name and is removed. It was `@deprecated` in favor of `'patch'`. No consumers in this repo's downstream apps (`sanity-io/sanity`, `sanity-io/canvas`) subscribe to it.

## Downstream impact

**Studio (`sanity-io/sanity`)**: Three `EventListenerPlugin` consumers — `PortableTextInput`, `StringInputPortableText`, `CommentInput`. All use exhaustive `switch (event.type)` with empty `default:`, so unknown event types are silently ignored. No runtime break and no TS break (the widened union still type-checks against an empty default). A real consequence worth flagging: these handlers now fire on every `insert.text` keystroke and `drag.dragover` mousemove, each hitting the switch and falling through to `default`. Sub-millisecond per event but a real hot-path cost on the most-used PT input. Worth measuring post-merge if anything feels sluggish.

**Canvas (`sanity-io/canvas`)**: Five `editor.on` subscriptions — all subscribe by specific type (`'selection'`, `'blurred'`, `'value changed'`). Unaffected. No exposure to the `'unset'` deprecation removal.

## Open questions

- The operation channel: `editor.on('patch', …)` carries Sanity patches. A `editor.on('operation', …)` channel carrying engine operations (`insert_node`, `unset`, `set`, etc.) would let plugins derive state like `listIndexMap` and `blockIndexMap` outside the engine without re-deriving from `value`. Follow-up.
- Naming: `drag.dragover` mirrors the DOM event name verbatim. Verbose but explicit. Worth a sanity check before locking the @public surface.

Spec: `/specs/shared-state-primitive.md` v5.

### DIFF
--- a/packages/editor/src/editor/create-editor.ts
+++ b/packages/editor/src/editor/create-editor.ts
@@ -14,7 +14,11 @@ import {createEditorDom} from './editor-dom'
 import type {EditorActor} from './editor-machine'
 import {editorMachine, rerouteExternalBehaviorEvent} from './editor-machine'
 import {mutationMachine, type MutationActor} from './mutation-machine'
-import {relayMachine, type RelayActor} from './relay-machine'
+import {
+  relayMachine,
+  type EditorEmittedEvent,
+  type RelayActor,
+} from './relay-machine'
 import {syncMachine, type SyncActor} from './sync-machine'
 
 export function createInternalEditor(config: EditorConfig): {
@@ -126,6 +130,11 @@ export function createInternalEditor(config: EditorConfig): {
           case 'value changed':
             listener(event)
             break
+          default:
+            // Behavior events (`insert`, `insert.text`, `drag.dragover`,
+            // `serialize`, etc.) are forwarded through the relay so consumers
+            // can subscribe by name: `editor.on('insert.text', listener)`.
+            listener(event)
         }
       })
 
@@ -279,6 +288,15 @@ function createActors(config: {
         case 'internal.patch':
           mutationActor.send({...event, type: 'patch'})
           break
+        case 'patches':
+          // PatchesEvent is an editor-machine internal event; not a behavior
+          // event and not forwarded to the relay.
+          break
+        default:
+          // Behavior events are emitted from the editor machine's
+          // `handle behavior event` action before the chain runs. Forward to
+          // the relay so `editor.on(behaviorEventType, ...)` fires.
+          config.relayActor.send(event as EditorEmittedEvent)
       }
     })
 
--- a/packages/editor/src/editor/editor-machine.ts
+++ b/packages/editor/src/editor/editor-machine.ts
@@ -398,47 +398,51 @@ export const editorMachine = setup({
         )
       }
     },
-    'handle behavior event': ({context, event, self}) => {
-      assertEvent(event, ['behavior event'])
+    'handle behavior event': enqueueActions(
+      ({context, event, self, enqueue}) => {
+        assertEvent(event, ['behavior event'])
 
-      try {
-        const behaviors = [...context.behaviors.values()].map(
-          (config) => config.behavior,
-        )
+        enqueue.emit(event.behaviorEvent)
 
-        performEvent({
-          mode: 'send',
-          behaviors,
-          remainingEventBehaviors: behaviors,
-          event: event.behaviorEvent,
-          editor: event.editor,
-          converters: event.editor.snapshot.context.converters,
-          keyGenerator: event.editor.snapshot.context.keyGenerator,
-          schema: event.editor.snapshot.context.schema,
-          readOnly: event.editor.snapshot.context.readOnly,
-          nativeEvent: event.nativeEvent,
-          sendBack: (eventSentBack) => {
-            if (eventSentBack.type === 'set drag ghost') {
-              self.send(eventSentBack)
-              return
-            }
+        try {
+          const behaviors = [...context.behaviors.values()].map(
+            (config) => config.behavior,
+          )
 
-            self.send(
-              rerouteExternalBehaviorEvent({
-                event: eventSentBack,
-                editorEngine: event.editor,
-              }),
-            )
-          },
-        })
-      } catch (error) {
-        console.error(
-          new Error(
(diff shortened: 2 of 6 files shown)


packages/editor/src/editor/create-editor.ts
```diff
@@ -126,6 +130,11 @@ export function createInternalEditor(config: EditorConfig): {
           case 'value changed':
             listener(event)
             break
+          default:
+            // Behavior events (`insert`, `insert.text`, `drag.dragover`,
```
christianhg, line 134: This new `default:` doing `listener(event)` matches every named case above it that also does `listener(event); break`. The whole switch can collapse to a single `listener(event)` call (drop the wrapping `switch (event.type) { ... }`). The named cases were a guard back when the relay's union didn't include behavior events; with `BehaviorEvent` now in `EditorEmittedEvent` they're noise.

Optional: keep the switch with `default: listener(event)` only, drop the 12 case branches.

packages/editor/src/editor/create-editor.ts
```diff
@@ -279,6 +288,15 @@ function createActors(config: {
...
+          // PatchesEvent is an editor-machine internal event; not a behavior
+          // event and not forwarded to the relay.
+          break
+        default:
+          // Behavior events are emitted from the editor machine's
+          // `handle behavior event` action before the chain runs. Forward to
+          // the relay so `editor.on(behaviorEventType, ...)` fires.
+          config.relayActor.send(event as EditorEmittedEvent)
       }
     })
```
christianhg, line 301: The comment frames this as "PatchesEvent is internal" but the load-bearing reason we filter is that the relay's `EditorEmittedEvent` doesn't include `PatchesEvent` — sending it would be a type error at runtime. Worth rewording:

```ts
case 'patches':
  // The relay's EditorEmittedEvent doesn't include PatchesEvent — drop it.
  break
```

Same for the default-case comment two lines below: the examples list (`insert`, `insert.text`, `drag.dragover`, `serialize`) is filler. The type union is the documentation. Could shrink to one line.

packages/editor/src/editor/editor-machine.ts
```diff
@@ -398,47 +398,51 @@ export const editorMachine = setup({
...
-      assertEvent(event, ['behavior event'])
+    'handle behavior event': enqueueActions(
+      ({context, event, self, enqueue}) => {
+        assertEvent(event, ['behavior event'])
 
-      try {
-        const behaviors = [...context.behaviors.values()].map(
-          (config) => config.behavior,
-        )
+        enqueue.emit(event.behaviorEvent)
```
christianhg, line 405: `enqueue.emit(event.behaviorEvent)` runs before the try/catch around `performEvent`. That's the right semantic (observe intent regardless of swallowing — also captured in the PR body), but the placement is load-bearing and worth pinning with a one-liner so a future move-into-the-try doesn't quietly flip the semantic:

```ts
// Emit before performEvent so subscribers see the event even if the chain throws or swallows it.
enqueue.emit(event.behaviorEvent)
```

packages/editor/tests/editor-on-behavior-event.test.tsx
```diff
@@ -0,0 +1,161 @@
+import type {PortableTextBlock} from '@portabletext/schema'
+import {createTestKeyGenerator} from '@portabletext/test'
+import {describe, expect, test, vi} from 'vitest'
+import {EventListenerPlugin} from '../src/plugins'
+import {createTestEditor} from '../src/test/vitest'
+
+describe('editor.on behavior event', () => {
+  test('editor.on fires per behavior event by name', async () => {
+    const onInsertText = vi.fn()
+    const keyGenerator = createTestKeyGenerator()
```
christianhg, line 10: Three test-quality nits per /behavior conventions:

1. **`expect(observed).toContain('tap:?')` on line 81 weakens the assertion.** Should be deep equality on the full array:

```ts
expect(observed).toEqual(['tap:?'])
```

2. **Tests 1, 2, and 3 share the same `createTestEditor` setup with `initialValue` and the same `select` event.** Worth hoisting a `setup()` helper or sharing fixtures so the test bodies surface only what differs (the `editor.on` subscription + the send + the assertion).

3. **`vi.waitFor`** polls until the assertion passes. `enqueue.emit` and the relay forward synchronously, so once `editor.send` returns the event has already been emitted — these can assert synchronously without `waitFor`. Tighter test, faster execution, and surfaces async surprises if they ever appear.

PR #2768, opened 2026-06-10
refactor: replace `editor.apply` wrappers with an operation channel

christianhg:
Operations were observed by wrapping `editor.apply`: update-value, patches and history each reassigned `apply` in `engine-plugins`, `withDOM` wrapped it for pending-state transforms, and the mutation and range-decorations machines wrapped it again on actor start. Ordering lived implicitly in the wrapper nesting, suppression in closure state, and nothing else could observe operations without adding another wrapper. The actor-based wrappers also restored `originalApply` on stop, clobbering any wrapper added after them.

Core `apply` now publishes every operation to a single channel (`engine/core/operation-channel.ts`). `subscribeToOperations` registers ordered, synchronous `before`/`after` listeners. The `OperationEvent` carries the operation, `beforeValue`/`beforeSelection` references captured at apply entry (operations replace these immutably, so the references stay pre-apply), the engine flags snapshotted at entry, and an `origin` derived from them. Normalization fixes re-enter `apply`, so a fix's events fire nested between the triggering operation's `before` and `after` events.

Each wrapper becomes a subscriber: `withDOM`'s pending-state transforms move to a `before`/`after` pair, index-map rebuilding to `subscriber.update-value`, patch generation to
`subscriber.patch-generation` (remote patch application split out into `remote-patches`), undo-step capture to `subscriber.history`, and the typing and range-decoration listeners to `before`-phase subscriptions that return their unsubscribe.

Net behavior unchanged; `editor.apply` is always the core apply.


christianhg:
Measured the suspected perf regression with `tests/performance.test.tsx` (chromium, 5 runs per variant, per-scenario medians, idle machine):

| Scenario | `main` | branch | branch + `1758f3b37` |
|---|---|---|---|
| 1000 blocks (empty) | 252.8ms | 252.6ms | 253.5ms |
| 1000 blocks before | 205.1ms | 206.9ms | 208.5ms |
| 1000 blocks after | 174.4ms | 173.5ms | 179.0ms |
| 1000 tables (empty) | 680.3ms | 674.3ms | 674.5ms |
| 1000 tables before | 575.6ms | 574.3ms | 576.9ms |
| 1000 tables after | 649.2ms | 641.1ms | 666.8ms |
| nested list 30× (total) | 1625ms | 1632ms | 1621ms |

The branch is at parity with `main` — every delta is smaller than the run-to-run spread within a variant. The channel's per-operation overhead (one event object + two listener-array copies) is real but sits well below the per-operation cost of index-map rebuilds and React commits.

`1758f3b37` removes that overhead anyway on principle (the apply path runs per keystroke): copy-on-write listener lists so emits iterate without copying, no event allocation when nothing is subscribed, the normalization debug listener only subscribed when `debug.normalization` is enabled, and patch generation bailing on `isPatching` before any per-operation work — the latter zeroes this subscriber's cost during remote patch application and value sync.

PR #2774, opened 2026-06-10
fix(sanity-bridge): avoid combinatorial blow-up converting mutually-embedding types

tfreeborough:
While upgrading our studio from v4 to v5 we hit a complete freeze: opening any document with a portable text field would pin the main thread until the tab crashed. No errors, nothing in the console. We eventually traced it to `sanitySchemaToPortableTextSchema`.

The walk added in #2630 tracks ancestors per branch, so it only stops when a type repeats within its own chain. Our schema has ~30 block object types, and about a dozen of them embed the shared `blockContent` array again (accordions, quotes, cards and so on). With that shape the walk ends up visiting every possible path through the type graph, which grows factorially with the number of block objects. For us that was ~2.1s per conversion — and since `PortableTextInput` converts on render, the studio just locked up. A slightly more connected schema never finishes at all.

The fix relies on the fact that `Schema.compile` produces one canonical instance per named type, so every position that embeds the shared array resolves to the same member instances. I added a small memo (scoped to one conversion call) keyed by the compiled type instance: each instance gets expanded once and later positions reuse the result, which makes the walk linear in the size of the compiled schema.

A few things I was careful about:

- The memo key is the instance, not the name. Same-named inline declarations with different fields compile to distinct instances, so they keep their own shapes — the cases in `same-name-objects.test.ts` are unaffected.
- Cycle detection via ancestor names is unchanged, and only completed expansions are memoized, so cycles still produce the same bare-reference stubs as before.
- All 15 existing tests pass without modification. The recursion fixtures declare their members inline at each call site, so they compile to distinct instances and the memo never fires there — output is byte-identical.

Numbers: the new regression test (12 block objects sharing a named array type) runs in ~5ms with the fix. Without it, vitest times out and has to kill the worker because the walk never returns. Against our production schema the conversion goes from 2,122ms to 1ms with identical output.

One semantic note worth flagging: a memoized expansion is computed under the ancestor chain of wherever it was first reached. If the same instance shows up later under a different chain, it reuses that first expansion, which can put a cycle stub where the old walk would have inlined one level deeper (or the other way around). Stubs resolve by name against `blockObjects`/`inlineObjects`, so no information is lost — and reproducing the old path-dependent output exactly isn't really an option, since that output is itself combinatorially large. No existing test pins a case where the difference is observable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/sanity-bridge-combinatorial-walk.md
+++ b/.changeset/sanity-bridge-combinatorial-walk.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/sanity-bridge': patch
+---
+
+fix: avoid combinatorial blow-up converting mutually-embedding types
--- a/packages/sanity-bridge/src/mutually-embedding-types-performance.test.ts
+++ b/packages/sanity-bridge/src/mutually-embedding-types-performance.test.ts
@@ -1,0 +1,71 @@
+import {Schema as SanitySchema} from '@sanity/schema'
+import {defineArrayMember, defineField, defineType} from '@sanity/types'
+import {describe, expect, test} from 'vitest'
+import {sanitySchemaToPortableTextSchema} from './sanity-schema-to-portable-text-schema'
+
+/**
+ * Regression test for a combinatorial blow-up in the deep schema walk.
+ *
+ * When many block-object types embed the same named Portable Text array
+ * type (a common pattern in real studios: a shared `blockContent` array
+ * reused as a field inside accordions, quotes, cards, ...), Sanity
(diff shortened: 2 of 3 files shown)


tfreeborough:
@christianhg maybe one for you to look at, we've pinned our version to 3.0.0 for now with Sanity studio v5.X.X

christianhg (approved):
Thanks a lot for this, @tfreeborough ❤️

tfreeborough:
Thanks @christianhg !

PR #2801, opened 2026-06-17
fix(perf): stop per-operation O(n) rescans during bulk structural edits

christianhg:
Inserting many blocks at once, or editing inside a large document, did work whose cost grew with the size of the whole document on every operation, so a bulk insert of N blocks spent O(N^2) in the engine before React ever rendered. Two independent per-operation rescans were responsible; this removes both. Each is its own commit.

## `listIndexMap` rebuilt on every operation

`listIndexMap` (the source of `data-list-index` on rendered list items) is derived from the whole value, because list-item numbering depends on block adjacency. The update-value subscriber rebuilt it from scratch after every operation whose path was at most two segments deep, walking all root blocks each time. A batch of N root operations therefore did N full walks, O(N^2), even though nothing reads the map until the next render, and the `path.length <= 2` gate silently assumed list numbering can only live at the document root.

The only reader is the text-block renderer. Structural operations now just mark the map dirty (on any path, dropping the depth assumption) and `getListIndexMap` rebuilds it on the next read, so a batch collapses to a single rebuild per render and the map stays correct for structural changes at any depth.

## Duplicate-`_key` normalization scanned siblings per node

`normalizeNode`'s duplicate-key fix resolved one node at a time: for each dirty node it fetched the node's siblings and scanned them for a collision, and at the root it rebuilt a wrapper array (`value.map(...)` plus a spread) per node. Normalizing a sibling group of n nodes was O(n^2) and, at the root, allocated O(n) throwaway arrays per node, so a bulk insert of pre-keyed blocks spent O(n^2) here finding nothing, because the keys were already unique.

A sibling group's key-uniqueness only changes when its membership changes, which mints a fresh parent reference. Nested groups now cache the verified-unique verdict in a `WeakMap` keyed by parent node and re-scan only when that reference changes. The root group has no parent node (its container is the editor, whose reference never changes), so its verdict lives in `editor.rootKeysVerifiedUnique`, set after a clean scan and cleared by the op stream only when an operation can add, remove, or re-key a direct root child (`affectsRootMembership`). The scan itself reads the raw child array straight off the parent node (the field segment is already in the node's path) instead of re-resolving children through the schema, and short-circuits groups of one child, which can never collide. The membership condition is exact, not a depth heuristic, and over-reports rather than under-reports, so a missed duplicate is impossible.

## Blast radius

Net behavior is unchanged: rendered `data-list-index` values are identical, and duplicate keys are still detected and the same later occurrence re-keyed in the same order. Only when the work happens changed (per-operation to per-read for the list map; once-per-group instead of once-per-node for the key scan). The full unit suite (801) and the chromium (1645) and webkit (1624) browser suites pass; `tests/focus.test.tsx` fails only in firefox and fails identically on `main` (it passes in isolation; a pre-existing focus-stealing flake), so it is untouched by this change.

Chromium, headless, medians of 3 runs, `tests/performance.test.tsx`:

| scenario | before | after |
|---|---|---|
| insert 1000 blocks (empty editor) | 244ms | 210ms |
| insert 1000 blocks before a block | 182ms | 149ms |
| insert 1000 blocks after a block | 159ms | 130ms |
| insert 1000 tables (empty editor) | 723ms | 624ms |
| insert 1000 tables before a block | 739ms | 635ms |
| insert 1000 tables after a block | 767ms | 591ms |

The remaining time is dominated by React mounting the 1000 components, which is inherent and untouched.

### DIFF
--- a/.changeset/lazy-list-index-map.md
+++ b/.changeset/lazy-list-index-map.md
@@ -1,0 +1,10 @@
+---
+'@portabletext/editor': patch
+---
+
+fix(perf): rebuild the list-index map lazily on read instead of per operation
+
+The map behind `data-list-index` on list items is now rebuilt once when the
+renderer reads it, rather than recomputed after every structural operation.
+Inserting or moving many blocks in one go no longer does per-block work that
+grows with document size, and the rendered index values are unchanged.
--- a/.changeset/verified-unique-key-scan.md
+++ b/.changeset/verified-unique-key-scan.md
@@ -1,0 +1,10 @@
+---
+'@portabletext/editor': patch
+---
+
+fix(perf): scan each sibling group for duplicate keys once instead of per node
+
+Duplicate-`_key` normalization no longer slows down quadratically on large
+sibling groups: a group's keys are checked once and the verdict reused until
+the group changes, so bulk-inserting blocks or editing inside big documents
+is faster. Duplicate keys are still detected and fixed the same way.
--- a/packages/editor/src/editor/create-editor-engine.tsx
+++ b/packages/editor/src/editor/create-editor-engine.tsx
@@ -69,6 +69,7 @@ export function createEditorEngine(
   editor.history = {undos: [], redos: []}
 
   editor.listIndexMap = new Map<string, number>()
+  editor.listIndexMapDirty = false
   editor.remotePatches = []
   editor.undoStepId = undefined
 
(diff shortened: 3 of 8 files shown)


christianhg:
Superseded by two isolated PRs: #2802 (lazy list-index map) and #2803 (verified-unique sibling-group key scan). The duplicate-key change there is reworked to use a serialized-path set invalidated by the op stream rather than a `WeakMap` keyed on the parent node, so it is uniform at every depth and an edit deep inside a large container no longer re-scans its ancestors.

PR #2813, opened 2026-06-18
feat: every container around the selection is focused

christianhg:
A container's `focused` prop in a `defineContainer` render callback now reports `true` for every container ancestor that fully contains the current selection, matching the symmetry `selected` has had since v7.

Previously the focused-container loop in `get-selection-state` stopped at the first hit — only the innermost container reported `focused: true`. An outer container chrome (the table around a focused cell, the code block around a focused line) was silently saying it wasn't focused even though the caret was inside it. Consumers wiring container chrome to focused state either accepted the lie or rebuilt the ancestor walk themselves.

### The rule

A container reports `focused: true` when the **entire current selection is contained within it**.

For a collapsed caret, every container ancestor satisfies this — the caret is a point, every ancestor contains that point, every ancestor reports focused.

For an expanded selection, only containers whose subtree contains both endpoints satisfy this. Computed as the intersection of the anchor's and focus's container ancestors — a container holds both endpoints iff it is an ancestor of both.

### Behaviour delta

| Scenario | Before | After |
| --- | --- | --- |
| Caret in a code-block line | code block: `false`, line: `true` | code block: `true`, line: `true` |
| Caret in a table cell paragraph | table/row/cell: `false`, paragraph: `true` | all four: `true` |
| Selection across two cells in one row | nothing focused | row + table focused; cells not |
| Selection across two rows | nothing focused | table focused; rows + cells not |
| Selection across two top-level paragraphs | nothing focused | unchanged — no common container |

No existing `true` becomes `false`. All deltas are `false → true` for legitimate "the user is editing inside me" cases.

### Implementation

`SelectionState.focusedContainerPath: string | undefined` becomes `focusedContainerPaths: Set<string>`, mirroring `selectedContainerPaths`. `useIsFocusedContainer` keeps its signature and contract (subscribe to "is this container holding the focus"); the implementation swaps the scalar equality check for `.has(serializedPath)`. The store-and-subscribe shape in `selection-state-context.tsx` is unchanged — same per-slice `useSyncExternalStore` mechanism, same re-render surface.

The collapsed branch walks ancestors once. The expanded branch walks anchor ancestors and focus ancestors and intersects — no new traversal util needed.

### Tests

- `src/editor/get-selection-state.test.ts` — covers every case in the matrix above plus backward selection. Each test asserts the FULL `focusedContainerPaths` set via `toEqual(new Set([...]))`, not spot checks.
- `tests/container-render-focused-selected.test.tsx` — updates the existing browser tests where they encoded the innermost-only rule. The expanded-across-cells test keeps its expectations: cells still report `focused: false` because neither cell contains both endpoints.

Targeted browser sweep: all container suites green (64/64).

### DIFF
--- a/.changeset/focused-ancestors-rule.md
+++ b/.changeset/focused-ancestors-rule.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': minor
+---
+
+feat: every container around the selection is focused
+
+A container's `focused` prop in a `defineContainer` render callback is now `true` for every container ancestor that fully contains the current selection, matching the symmetry that `selected` already had. Previously only the innermost container reported `focused: true`, so an outer container chrome (table around a focused cell, code block around a focused line) had no way to react to the caret being inside it.
+
+For a collapsed caret, every container ancestor reports focused. For an expanded selection, only containers whose subtree contains both endpoints report focused: a selection across two cells reports the row and table focused but not the cells.
--- a/packages/editor/src/editor/get-selection-state.test.ts
+++ b/packages/editor/src/editor/get-selection-state.test.ts
@@ -10,7 +10,7 @@ describe(getSelectionState.name, () => {
     expect(getSelectionState(testbed.snapshot, null)).toEqual({
       focusedLeafPath: undefined,
       selectedLeafPaths: new Set(),
-      focusedContainerPath: undefined,
+      focusedContainerPaths: new Set(),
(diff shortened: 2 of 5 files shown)


christianhg:
Closing as part of the open-PR cleanup. The branch preserves the implementation (June state, not re-verified against current `main`); tracked in Linear for revival, likely alongside the native tables field work.

PR #2814, opened 2026-06-18
feat: add unified input handling across desktop and mobile

christianhg:
## What

On Android, autocorrect did not work and every keystroke paid for a full DOM
parse-and-diff reconciliation. On desktop, typing a character immediately
after toggling a decorator at a collapsed cursor left a stray, undecorated
ghost copy of that character in the DOM:

```
type "foo", toggle bold (⌘B), type "bar"
  rendered: foobbar
  model:    foo + bold "bar"   (correct)
```

Both are symptoms of the same thing: the editor ran two separate input
pipelines. This replaces them with a single `InputManager` that handles input
the same way on every platform. It carries forward #2276, which had drifted
~475 commits behind its March base, rebuilt on top of current `main` so it can
land without re-litigating a rebase of that size.

## Design notes

The Android pipeline (`android-input-manager`, 946 lines) assumed every
`beforeinput` event was non-cancelable, so it never called `preventDefault()`:
it let the browser mutate the DOM and then reconstructed the model by diffing
the DOM on every keystroke. That is both slow and the reason autocorrect
misbehaved. In fact only IME composition is non-cancelable; ordinary typing on
Android is cancelable like it is everywhere else.

The desktop path had the opposite failure. For a single-character `insertText`
on a collapsed, non-zero-offset selection it took a native fast-path: instead
of `preventDefault()`-ing, it let the browser write the character into the
focus span's text node and deferred the model `insert.text` to the `input`
event. That is fine until the model insert diverges from where the browser
wrote. A pending decorator toggle routes the next insert into a new span, so
the browser's in-place edit stays in the old (focus) span while the model
grows a new one. The focus span's model node never changes, its memoized leaf
never re-renders, and the layout effect that would reset the stray text never
runs. The character ghosts.

`InputManager` (`src/dom/input-manager.ts`, wired by `use-input-manager.ts`)
branches on `event.cancelable` rather than on platform. When the event is
cancelable (desktop typing, delete, paste, Enter, and the bulk of Android
input) it calls `preventDefault()` and dispatches the corresponding behavior
event (`insert.text`, `delete.backward`, and so on) directly, so the browser
never touches the DOM and the model stays the single source of truth. When it
is not cancelable (composition, IME, mobile spellcheck) it snapshots the block
text, lets the browser mutate, then reads the mutated DOM and diffs it against
the snapshot (`change-detector.ts`) to recover the behavior event
(`change-to-behavior-event.ts`). The expensive parse-and-diff now runs only
for composition, not for every keystroke.

The load-bearing invariant is that cancelable input never mutates the DOM
ahead of the model. That is what removes the ghost class outright: there is no
native fast-path left, so there is no stray edit for a non-re-rendering leaf
to strand. `tests/decorator-toggle-no-ghost.test.tsx` pins it (type `foo`,
toggle bold, type `bar`, assert the rendered text equals the model `foobar`)
and is green with no special-casing. `editable.tsx` sheds ~500 lines as the
input logic moves into the manager.

Deleting the Android manager and the desktop fast-path removed the only
consumers of several vendored engine helpers, so they are removed here rather
than left as dead exports: the seven now-unused functions in
`engine/dom/utils/diff-text.ts` (the file keeps its `TextDiff` type and the
still-used `transformPending*` / `transformTextDiff` functions), the `DOMText`
value export in `dom.ts` whose "used in `dom-editor.ts`" comment was stale, the
`IS_IOS` / `IS_FIREFOX_LEGACY` / `IS_UC_MOBILE` / `IS_WECHATBROWSER` flags in
`environment.ts`, and `engine/path/is-descendant-path.ts`, which became
unreachable once `diff-text.ts` stopped importing it. `main` carries zero
unused exports; this keeps `knip` at zero.

Green on types, lint, knip, format, the unit suite (853), and the chromium
(1647) and webkit (1626) browser suites.

One firefox regression is open and is the blocker before this can be marked
ready: `tests/focus.test.tsx > Focusing on an empty editor` fails, the focus
event settling on `blurred` instead of `focused`. It passes on `main` in
firefox, so it is introduced here, and it is isolated to the
focus/selection-ensuring path in `editable.tsx`: that flush used to be gated
on `IS_ANDROID`, and the unified path drops the platform guard, so it now runs
on firefox, where focusing an empty contenteditable resolves the selection
differently. The empty-editor focus path needs reconciling with the unified
flush; under investigation.

Same change as #2276, rebuilt on current `main`, so #2276 should be closed in
favor of this. #2812 (the targeted `decoratorState` guard that suppresses the
ghost on the existing fast-path) becomes unnecessary once this lands, since
the fast-path is gone.


christianhg:
Closing as part of the open-PR cleanup, parked without prejudice: the unified input manager remains the direction, and this branch (`feat/unified-input-manager`, June state, not re-verified against current `main`) preserves the rebuilt implementation and its design notes. The rearchitecture effort owns the one-input-pipeline goal; revival is rebase and reopen.

PR #2829, opened 2026-06-19
fix(perf): recompute toolbar button state once per microtask, not per editor event

christianhg:
Selecting all and deleting in a large document freezes the main thread. A CPU profile of the deployed editor (select-all + delete) spends the majority of its active time not in the engine but in the toolbar: `editor.on('*')` → `isActiveAnnotation`/`isActiveDecorator`/... → `getSelectedSpans` → `getNodes` over the whole selection → `blockIndexMap.get` per node.

Every toolbar button (annotation, decorator, style, list, the annotation popover, and the shared disabled-state listener) subscribes to `editor.on('*')` and recomputes its active state on every editor event. Those selectors walk the current selection, so each recompute is O(selection). A single user action emits one event per underlying operation, so deleting a large selection emits an event per removed block while the selection is still large. The cost is O(buttons × operations × selection), i.e. quadratic in document size, and it runs synchronously, so the tab hangs.

`subscribeToEditorChange` subscribes to `'*'` the same way but coalesces a synchronous burst of events into a single trailing call on the next microtask. Each selector then runs once per burst instead of once per event, so a delete that emits N operation events recomputes each button once over the settled selection rather than N times over progressively-smaller selections. The settled state is what the toolbar displays, so the trailing call is the only one that matters; a pending microtask that fires after unsubscribe is dropped.

The only observable change is that a button's active state settles one microtask after a change instead of synchronously within the event. The displayed result is unchanged. Each listener still computes its initial state synchronously on mount, so first paint is unaffected. A unit test pins the contract: 100 synchronous events produce exactly one listener call, a second burst produces one more, and no call fires after unsubscribe.

This is the toolbar half of the select-all-delete hang. The engine itself also has O(n²) terms on that path, addressed separately by `editor-perf-delete-range-backwards` (reverse range removal) and `editor-perf-lazy-list-index` (deferred list-index rebuild); this PR is what removes the dominant cost in the profile.

### DIFF
--- a/.changeset/toolbar-coalesce-active-state.md
+++ b/.changeset/toolbar-coalesce-active-state.md
@@ -1,0 +1,10 @@
+---
+'@portabletext/toolbar': patch
+---
+
+fix(perf): recompute toolbar button state once per microtask, not per editor event
+
+Toolbar buttons no longer freeze the editor while recomputing their active
+state during large edits. Deleting or changing a big selection previously
+recomputed every button's active state on every underlying operation, each
+recompute scanning the whole selection; the state now settles once per change.
--- a/packages/toolbar/src/disable-listener.ts
+++ b/packages/toolbar/src/disable-listener.ts
@@ -1,5 +1,6 @@
 import type {Editor} from '@portabletext/editor'
 import {fromCallback, type AnyEventObject} from 'xstate'
+import {subscribeToEditorChange} from './subscribe-to-editor-change'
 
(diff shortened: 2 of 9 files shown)


christianhg:
Superseded by #2831, which merged the toolbar migration using the first-class `editor.on(..., {schedule: 'microtask'})` option instead of this branch's local `subscribeToEditorChange` helper.

PR #2851, opened 2026-06-24
feat(playground): draw drop indicators for container blocks

christianhg:
Dragging a callout, fact-box, or table around the playground gave no hint of where it would land. Move it over a paragraph and a thin line shows the drop edge; move it over another container and there is nothing. The drag still works, you just can't see the target.

The line is engine chrome that the new render pipeline opts out of. For legacy top-level blocks the engine tracks a drop position during the drag and `render.text-block.tsx` draws a 1px `DropIndicator` for it via `useElementDropPosition`. Blocks rendered through the `defineX` pipeline (every container and its nested text blocks) render no indicator on purpose: where a dragged block lands is pointer-driven UI, not document structure, so the engine leaves it to the consumer.

`@portabletext/plugin-dnd` is that consumer hook. Its `useDropPosition(path)` exposes the same position the engine computes, derived from the public `drag.*` behavior events, as `'start' | 'end' | undefined`. This mounts `DndProvider` inside `EditorProvider` (alongside the existing `ListIndexProvider`) and adds a shared `BlockDropIndicator` that reads the position for its path and draws the line: 1px `currentColor` at the block's top edge for a `'start'` drop, the bottom for `'end'`, matching the engine's `pt-drop-indicator` so container and top-level indicators look identical. It is a `<span>` so it stays valid markup inside a `<p>` text block, and `position: absolute` so it never reflows the text; the block element it renders into is `relative`.

```tsx
function BlockDropIndicator({path}: {path: Path}) {
  const position = useDropPosition(path)
  if (position === undefined) return null
  return <span style={{position: 'absolute', /* top or bottom: 0 */ borderTop: '1px solid currentColor'}} />
}
```

Putting that on every block type without copying it drove a consolidation. The callout, fact-box, and cell text-block renders each repeated the same shape: a list-item branch (re-emitting the `data-list-*` attributes the counter CSS keys off) and a `switch (node.style)` returning a styled element, differing only in which styles each container allows. That collapses into one `ContainerTextBlock` parameterized by a `{style: {tag, className}}` map, which subsumes the former `ListItemBlock` (the `data-list-*` re-emission moves into it unchanged) and renders the indicator in one place. The three top-level container elements render `BlockDropIndicator` directly.

List rendering output is unchanged: the same `data-list-*` attributes on the same wrapper, the same element and classNames per style, now with `position: relative` added (no offsets, so no layout shift) and the indicator as a trailing child. Row and cell `<tr>`/`<td>` get no direct indicator yet: table-internal absolute positioning is fragile and neither is drag-initiated, so their content blocks carry the coverage. The position tracking itself is pinned in the plugin's own `plugin.dnd.test.tsx`; the playground side is render glue, so it leans on that rather than restating it.


christianhg:
Closing as part of the open-PR cleanup. Playground-only change, branch preserved (June state, not re-verified against current `main`); tracked in Linear for revival.

PR #2863, opened 2026-06-26
feat(editor): select a container as a block-object via a spacer

christianhg:
A container (`defineContainer`, e.g. a code-block, callout, or table) can't be selected, deleted, or dragged as a unit. Clicking its chrome places a caret inside the content; there is no way to select the container itself or Backspace it. Void blocks get this for free, and the reason is narrow: a void is childless, so the hidden selection spacer the engine renders inside it is its one selectable position, and a click anywhere on the void's non-editable body routes to it. A container has an editable body, so a point on it drills to the first leaf, the editable body competes for every click, and the block-object keyboard pipelines gate on `isLeafObject`, which excludes editable containers.

This gives a container the same spacer, adapted for an editable body.

## Userland

`defineContainer`'s `render` now receives a `spacer`. Render it in the container's chrome and layer the editable content above it; clicking the chrome then selects the container, clicking the content places a caret. Omit the spacer and the container stays caret-only.

```tsx
defineContainer({
  type: 'code-block',
  arrayField: 'lines',
  render: ({attributes, spacer, children}) => (
    <pre {...attributes} className="relative">
      {spacer}
      <code className="relative z-10">{children}</code>
    </pre>
  ),
})
```

The second commit wires exactly this into the playground's code-block.

## Design notes

The spacer's presence is the only signal. Nothing registers a container as "selectable"; rendering the spacer is the opt-in, detected at runtime, the same way voids work. There is no `selectable` flag to keep in sync.

The spacer fills its region rather than being a zero-height anchor. A void's spacer can be `height:0` because the void's whole body is non-editable, so the spacer is the only thing a click can land on. A container's body is editable and competes for clicks, so the container spacer is `position:absolute; inset:0`: it occupies the chrome, and the consumer layers the editable content above it (`z-10`). A click on the chrome lands on the spacer, a click on the content lands on the content.

The DOM/model mapping mirrors voids and reuses the void spacer markers (`data-pt-spacer` / `data-pt-marks` / `data-pt-zero-width`); the engine tells a container's spacer apart from a void's, or from the spacers/zero-widths throughout the container's editable body, by its nearest `data-pt-block` being a `container`. `toSelectionPoint` resolves a point in such a spacer to `{path: container, offset: 0}` (void spacers fall through to the existing void handling). `toDOMPoint` renders a container selection back onto that container's own spacer, scoped past nested containers and void children. `onDOMSelectionChange` accepts the spacer as a selectable anchor (it may sit in non-editable chrome, so it is neither an editable target nor inside a void) and commits the already-resolved container point as-is.

That "as-is" is the crux, and it's why there is no flag. `resolveSelection` is left completely untouched: it still drills a container point to the first leaf, which is correct everywhere content gets positioned, `insert` `at` a container, a drop target, the selection that resolves after a delete. Removing that drill wholesale breaks exactly three things and they are all content-positioning (proven by `event.insert.block`, `event.drag`, and the `behavior-api` delete-then-insert scenario). So instead of teaching `resolveSelection` an exception, the spacer selection simply never goes through it: `onDOMSelectionChange` commits it directly. The block-object intent is born and consumed at the one place that knows, the spacer, and never travels as a flag through the event/operation layers.

While a container is selected, four behaviors act on it: delete backward/forward `unset` the container, ArrowUp/ArrowDown move the caret to the adjacent block. `unset` is used rather than `delete.block` so the whole container is removed without re-resolving through the drill.

`container-spacer.test.tsx` pins the four observable behaviors (spacer selects the container, body still carets, Backspace deletes it, ArrowDown navigates) with real DOM selection and keyboard. Full unit suite (809) and browser suite (1680) pass with no regressions.

## Not covered

How a spacer-less container is treated is unchanged (caret-only). The chrome layout (which regions select vs caret, the stacking) is the consumer's call; the playground demonstrates one. Esc-to-select is separate.

---

> [!NOTE]
> **Medium Risk**
> Touches selection sync, DOM mapping, and keyboard behaviors in the editor core; opt-in via spacer limits blast radius, but incorrect handling could affect caret placement or delete/navigation around containers.
> 
> **Overview**
> `defineContainer` **render** callbacks now receive an optional **`spacer`** element. Placing it in the container chrome (with editable **`children`** layered above) lets users select the whole container like a block object; omitting **`spacer`** keeps today’s caret-only behavior.
> 
> The engine wires **DOM ↔ model** for container spacers: **`toSelectionPoint`** / **`toDOMPoint`**, **`onDOMSelectionChange`** treats container spacers as selectable anchors, and commits container selections via **`applySelect`** so **`resolveSelection`** does not drill into the first inner leaf. New core behaviors **`unset`** the container on Backspace/Delete and move selection with ArrowUp/ArrowDown when the container path is selected.
> 
> The playground **code-block** plugin demonstrates **`spacer`** + **`z-10`** stacking; **`container-spacer.test.tsx`** covers selection, caret-in-body, delete, and navigation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7898bb978e4678408c0d2e708b1a4c55f3f70cbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


christianhg:
Closing as part of the open-PR cleanup. The branch preserves the spacer-based container selection work with its tests (June state, not re-verified against current `main`); tracked in Linear for revival.

PR #2878, opened 2026-06-29
fix: validate `getSibling`'s `blockIndexMap` index against the value

christianhg:
Found by the traversal audit that followed the `getNode` path-contract fix (#2876, EDEX-1425), same family: a util over-trusting `blockIndexMap` without the value fallback the rest of the traversal layer carries.

`getSibling` looked up the current node's index with `blockIndexMap.get(serializePath(path))` and trusted it blindly, then indexed the value-derived sibling array (from `getChildren`) with it. `blockIndexMap` can **miss** (an unkeyed transient node) or **disagree** with the traversed value (a snapshot that pairs the live map with a pre-apply value, e.g. `textPatch`):

- On a **miss**, the lookup returned `undefined`, so `getSibling` reported *no sibling* where one exists.
- On a **disagreement**, the stale map index pointed at the wrong slot in the value-derived array, returning the *wrong* sibling.

`getNode`, `getChildren`, and `getAncestors` all take the map as a **fast path** and fall back to a value scan when it misses or disagrees; `getSibling` was the lone holdout trusting it unconditionally. It now validates the mapped index against the resolved `children` (does the child at that index carry the path's last keyed `_key`?) and scans only when it doesn't.

The O(1) lookup is preserved in the common case; the O(n) scan runs only on the stale-map snapshots. `children` was already materialized by `getChildren` (itself O(n)) to return the sibling node + path, so the fast path adds no asymptotic cost over what the function already did.

TDD: the new test (`resolves the sibling from the value when the blockIndexMap misses`) pairs the testbed value with an empty map and asserts next/previous/nested-span siblings still resolve. It fails on `main` (returns `undefined`) and passes with the fix; the existing 21 cases are unchanged.

Net behavior is unchanged when the map and value agree. `816` editor unit tests pass; types, lint, knip green.

### DIFF
--- a/.changeset/get-sibling-value-index.md
+++ b/.changeset/get-sibling-value-index.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: validate `getSibling`'s `blockIndexMap` index against the value
+
+`getSibling` still uses the block index map as an O(1) fast path, but now validates the index it returns against the resolved sibling list and falls back to a value scan when the map is missing or stale, an as-yet-unkeyed node, or a snapshot that pairs the live map with a pre-apply value. Previously it trusted the map blindly and returned `undefined` (reporting no sibling where one exists) or, when the map disagreed with the value, the wrong neighbour. This matches how `getNode`, `getChildren`, and `getAncestors` already treat the map.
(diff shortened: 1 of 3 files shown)


christianhg:
Superseded: #2894 shipped the same verify-then-fallback fix (plus numeric-segment support), and #2915 then consolidated the resolution into the shared `resolveChildEntryIndex`, so `getSibling` no longer contains the inline lookup this patches. The miss/disagreement holes are pinned by #2894's tests through what is now the single shared resolution path.

PR #2899, opened 2026-07-02
fix: skip leafless subtrees when normalizing DOM points

christianhg:
Triple-clicking a table in the playground shows a whole-table selection on screen while the editor's model keeps a stale collapsed caret, so Delete and typing act on something other than what's visibly selected. Clicking near the table's edges similarly parks carets that never map into the model (EDEX-1118 reported the same family from the deployed playground in June: typing after such a click injects content into the last cell, then below the table).

The diagnosis: `getEditableChildAndIndex` treats any element with children as content. A table's `<colgroup>` has children (the `<col>`s) but no editable content, so `normalizeDOMPoint`'s descent walks into it and dead-ends: the resulting point on a `<col>` resolves to nothing in `toSelectionPoint`, `suppressThrow: true` swallows the failure, and the whole `selectionchange` sync silently skips, leaving the model stale. Any leafless subtree inside a container render triggers it; `<colgroup>` is just the canonical case every table renders.

The fix extends the child-seeking skip condition: elements whose subtree holds no editable content (no text, no rendered block/inline object) are skipped with backtracking, exactly as `contenteditable="false"` elements already are. The predicate is deliberately conservative: zero-width leaves keep matching through their text content and void objects through their rendered attributes, so void-adjacent positions behave exactly as before.

Two tests pin it on a container table rendered with a plain `<colgroup>` (no attribute tricks), both driven through the real `selectionchange` sync: an element-level range spanning the table must map to first-cell start to last-cell end, and a collapsed element-level point before the colgroup must map into the first cell. Both fail without the fix (one maps wrong, one maps to `null`), verified red. Green on chromium, firefox, and webkit; the full chromium editor browser suite passes 1690/1690.

Follows #2898 in the same subsystem: that one stopped the validator from destroying equivalent DOM selections, this one makes the mapping produce the right answer in the first place. The remaining piece of the cluster is EDEX-1512 (canonicalizing collapsed carets parked on structural elements).

### DIFF
--- a/.changeset/normalize-dom-point-leafless-subtrees.md
+++ b/.changeset/normalize-dom-point-leafless-subtrees.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: selection mapping no longer dead-ends in leafless subtrees
+
+Selecting across a table rendered with a `<colgroup>` (for example via triple-click, or clicking near the table's edges) could silently fail to update the editor's selection, leaving it stale or collapsed while the screen showed a larger selection. DOM positions that normalize into an element holding no editable content now skip to the nearest sibling with content instead, so element-level selections map to what they visually cover.
--- a/packages/editor/src/engine/dom/utils/dom.ts
+++ b/packages/editor/src/engine/dom/utils/dom.ts
@@ -160,12 +160,16 @@ const getEditableChildAndIndex = (
   let triedForward = false
   let triedBackward = false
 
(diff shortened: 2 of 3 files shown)


christianhg:
Closing as part of the open-PR cleanup, with the work preserved and verified: rebased onto today's `main`, the bug proven alive first (both tests red on `main`), and this PR's tests plus the selection-validation suite green on the branch. The full monorepo gauntlet was not completed before parking; run it on revival. Tracked in Linear.

PR #2905, opened 2026-07-03
feat: address `serialize` events with `at` and `fragment`

christianhg:
The serialize family had no addressing concept: `serialize` and `serialize.data` always serialize the live selection, so anything that needs to serialize different content has to work around the events rather than through them. Core itself is the first offender: the abstract `serialize.data` behavior pattern-matches `drag.dragstart` origins and doctors the snapshot with the grabbed selection before calling converters, a special case that exists precisely because the event cannot carry its subject. The table plugin's rectangle copy (#2903) generalized that workaround, calling converters directly from its guard with a doctored snapshot, which makes its interception a terminal handler: the rectangle never exists as an observable event, so an app's own serialize behaviors either never see rectangle copies or, if registered first, serialize the wrong cells.

Both events now carry the serialization subject along its two independent axes, each optional and each meaningful alone. `at` serializes a different range of the document, reusing the addressing convention every other synthetic event already has; drag serialization moves onto it, and the origin-type pattern-match in `serialize.data` is deleted, so the diff removes a hack while adding the API. `fragment` serializes different content entirely, standalone blocks whose range defaults to the fragment's full span (the block edge points descend into containers, so a table fragment spans its first cell's first leaf to its last cell's last leaf). An earlier iteration shipped one required pair, `fragment: {value, selection}`; review found that shape to be the table plugin's own mechanics leaking into the API: drag only needs a range, content overrides only need blocks, and nobody needs to hand-compute leaf paths into their own synthetic value. The decomposition fell out of asking what each consumer actually holds.

The abstract `serialize.data` behavior resolves both fields into the snapshot it hands converters (`value = fragment ?? context.value`, `selection = at ?? fullSpan(fragment) ?? context.selection`). Converters change zero, they keep reading `snapshot.context.value` and `snapshot.context.selection` exactly as before. Interceptors become re-addressers instead of terminal handlers, subject-presence is the loop guard:

```ts
defineBehavior({
  on: 'serialize',
  guard: ({snapshot, event}) => {
    if (event.fragment) {
      return false
    }
    const table = sliceToRectangle(snapshot)
    return table ? {fragment: [table]} : false
  },
  actions: [({event}, {fragment}) => [raise({...event, fragment})]],
})
```

On honesty about consumers: the `at` axis has a non-table consumer today, drag itself. The `fragment` axis has exactly one consumer today, the table plugin; its justification is structural rather than a consumer list, without a content override, any plugin whose copyable unit is not expressible as a selection is forced back into terminal handlers, which is the composition failure this PR exists to prevent.

`tests/event.serialize.test.tsx` pins each axis separately: an `at` re-raise puts the addressed range on the clipboard, and a bare `fragment` re-raise (no `at`) serializes content that exists nowhere in the document, full-span by default. The dragstart unification is pinned by the existing drag serialization tests passing unchanged. Full suite: 1695 chromium tests plus the unit project, lint/types/knip/format green.

#2903 stacks on this: its interception collapses to the re-raise above.

### DIFF
--- a/.changeset/serialize-at-and-fragment.md
+++ b/.changeset/serialize-at-and-fragment.md
@@ -1,0 +1,23 @@
+---
+'@portabletext/editor': minor
+---
+
+feat: address `serialize` events with `at` and `fragment`
+
+`serialize` and `serialize.data` gain two optional fields naming the serialization subject: `at` serializes a different range of the document (the same addressing convention as the other synthetic events), and `fragment` serializes different content entirely, with the range defaulting to the fragment's full span. A behavior can re-raise a serialize event with a corrected subject and let the rest of the pipeline, including custom converters, handle it:
+
+```ts
+defineBehavior({
+  on: 'serialize',
+  guard: ({snapshot, event}) => {
+    if (event.fragment) {
+      return false
+    }
+    const table = sliceToRectangle(snapshot)
+    return table ? {fragment: [table]} : false
+  },
+  actions: [({event}, {fragment}) => [raise({...event, fragment})]],
+})
+```
+
+Converters are unchanged: the subject is resolved into the snapshot they receive, so they keep reading `snapshot.context.value` and `snapshot.context.selection` as before.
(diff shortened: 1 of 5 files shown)


christianhg:
Pausing this: the override shape went through three iterations in review (required pair, decomposed optional fields, exclusivity and mode questions) without settling, which is the signal it is not ready to become permanent event surface. The table plugin keeps its plugin-local doctored-snapshot interception (#2903), deletable when this lands. Design state and the decision trail live on the ticket; the branch stays.

PR #2906, opened 2026-07-03
fix(perf): render the root block list in memoized chunks

christianhg:
Typing in a large document paid a React cost proportional to the total block count: the whole document renders as one flat array under a single `Children` component, so every content edit re-ran the root `useChildren` (rebuilding all n JSX elements, ~12.3ms per keystroke at 8k blocks) and forced React to reconcile all n child positions (fresh props objects per rebuild force the `MemoizedElement` comparator per child before it can bail). The per-node memo chain held, only the edited block's components actually re-rendered, but the rebuild and reconcile above it were O(n) per keystroke. This is the flat-root term from the render-architecture review, the dominant per-keystroke cost after #2893 and #2900, and the classic Slate scaling wall: upstream Slate grew experimental chunking for exactly this, and this codebase removed its inherited copy as unused (`58d4eb11b`), so this is a fresh root-only design rather than a revert.

The root children now render as contiguous chunks. `editor.rootChunks` is maintained per operation by `transformRootChunks` (`internal-utils/root-chunks.ts`), which follows the `transformBlockIndexMap` pattern exactly: a pure per-op transform that resolves the operation's root index against the pre-op `blockIndexMap` (verified against the chunks, scan fallback for unmaintained maps, full-rebuild fallback for unresolvable shapes), rebuilds only the owning chunk's block array, splits chunks above 200 blocks, and drops empty ones. Chunks never merge, a deliberate ceiling marked in the code: scattered deletions can accumulate small chunks and degrade the dispatch back toward the flat behavior in the pathological limit, and merge-with-neighbor is the upgrade if chunk counts ever show up in a profile. One ordering subtlety carries the correctness: every op except `set.selection` replaces the owning root block's reference (the tree updates immutably), so the transform must see text ops too and runs before the update-value subscriber's text-op early return. The oracle test pins `flatten(chunks) === value` across a 400-step fuzzed op sequence plus explicit identity-reuse, split, drop, map-miss, and numeric-path cases. The unit fuzz simulates op semantics by hand, so a browser integration test closes the loop against the real engine: it asserts `flatten(editor.rootChunks)` equals the value by per-block reference equality through a mixed session (bulk insert across chunk boundaries, real typing, splits, backspace merges, undos, select-all delete across every chunk, undo of the delete).

On the render side, the root `useChildren` dispatches to `MemoizedChunk` components keyed by chunk id, each re-entering `useChildren` with an explicit `chunkBlocks` slice. `splitDecorationsByChild` accepts the same override, so every chunk resolves the full root decorations against its own key-to-index map and decorations addressing blocks outside the chunk fall away naturally, no separate per-chunk decoration splitting exists. The chunk memo bails on chunk identity (the transform's contract: a chunk object is replaced exactly when its own blocks changed), render-prop identity, and decoration content equality, content rather than identity because the editable recreates the root decorations array every render. Chunks render as fragments, so the contenteditable DOM is byte-identical to before: selection, IME, `toDOMNode`, and `RestoreDOM` are untouched, which is the property that makes this a contained render refactor instead of a risky editor feature.

One regression the suite caught and the fix it forced: chunk-level render decisions (container resolution, pipeline membership) read the registration maps, and the chunk memo blocks the root re-render that used to refresh them, `container-normalization.test.tsx` failed on unmounting a `ContainerPlugin`. Chunks therefore subscribe to the `useRegistrationsSelector` channel from #2900, which fires only on renderer register/unregister; registrations are config-time, so the subscription is off the hot path. Everything else that used to reach blocks through root re-renders arrives via contexts, which pierce the memo.

React-side cost per keystroke (chromium, warmed and settled, medians): root dispatch plus one chunk body is 0.87ms at 1k blocks and 1.27ms at 8k, versus 0.85ms and 12.3ms for the flat map, size-independent where it was linear, with exactly one chunk body re-running per keystroke. Probe wall time per keystroke at 8k dropped from ~61ms to ~23.5ms. Full suite green: 848 unit including the new transform oracle, 1694 chromium. A second commit adds a typing-in-a-large-document baseline to the performance test file.

### DIFF
--- a/.changeset/chunked-root-children.md
+++ b/.changeset/chunked-root-children.md
@@ -1,0 +1,12 @@
+---
+'@portabletext/editor': patch
+---
+
+fix(perf): render the root block list in memoized chunks
+
+Typing and editing in large documents no longer pays a React
+reconciliation cost proportional to the total block count. The root
+block list now renders in contiguous chunks whose components skip
+re-rendering unless one of their own blocks changed, so a keystroke
+rebuilds and reconciles one chunk instead of the whole document. The
+rendered DOM is unchanged, chunks render as fragments.
--- a/packages/editor/src/editor/create-editor-engine.tsx
+++ b/packages/editor/src/editor/create-editor-engine.tsx
@@ -5,6 +5,7 @@ import {BlockIndexMap} from '../internal-utils/block-index-map'
 import {buildIndexMaps} from '../internal-utils/build-index-maps'
 import {createPlaceholderBlock} from '../internal-utils/create-placeholder-block'
 import {debug} from '../internal-utils/debug'
+import {buildRootChunks} from '../internal-utils/root-chunks'
 import type {PortableTextEditorEngine} from '../types/editor-engine'
 import type {EditorActor} from './editor-machine'
 import {setupRemotePatches} from './remote-patches'
@@ -104,6 +105,8 @@ export function createEditorEngine(
     subscriptions: config.subscriptions,
     editor: editorEngine,
   })
+
+  editor.rootChunks = buildRootChunks(editor.snapshot.context.value)
 
   buildIndexMaps(
     {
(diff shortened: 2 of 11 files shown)


christianhg:
Closing as part of the open-PR cleanup. State at closure: rebased onto today's `main`, full monorepo suite green, perf win re-verified (145.8ms -> 80.3ms per 20 keystrokes at 1000 blocks). One blocking review finding is recorded on the ticket: the chunk-slice decoration fallbacks paint chunks outside a decoration's range. Branch `chunked-root-children` preserves everything.

PR #2931, opened 2026-07-07
fix: validate `insert.block` sibling placements against the parent array's schema

christianhg:
While hardening table clicks in #2910 we found that `insert.block` with `{_type: 'block'}` addressed at a table row's path and `placement: 'before'` writes the text block into `rows`, schema-invalid data that nothing catches downstream. The click behaviors now guard against raising that event, but the operation itself kept writing when pointed there directly.

The operation already validates position-aware: it derives a destination from the operation's `at`, resolves `getPathSubSchema` there, and no-ops blocks the schema rejects. The defect is a scope disagreement specific to sibling placements. `resolveTarget` lands `before`/`after` blocks as siblings of the enclosing block at the destination, inside its parent array, while the validation scope derived from the destination path's own sub-schema view. The two only differ when the destination points at a container member: a point at a table row resolves toward the cell content view, which accepts text blocks, while the sibling insertion lands in `rows`, which does not.

Sibling placements now resolve their validation scope from the enclosing block's parent array. Acceptance is an explicit membership check against the enclosing container's `of`, and the parse then runs against that array's sub-schema so fields validate in the right scope. The membership check cannot be delegated to the parse: a `Schema` structurally always carries a text-block type, so a sub-schema alone can never reject `{_type: 'block'}`, which is the same reason `fitBlocksToDestination` layers its own accept-list checks over the paste pipeline. Root-level destinations are untouched, the root accepts everything the schema declares, and `placement: 'auto'` keeps the inner-view derivation since its landing logic (replace, merge, first-child descent) genuinely targets the destination's inside.

`tests/insert-block-placement-scope.test.tsx` pins four corners: the disallowed sibling no-ops (red before the fix, with the block landing inside `rows`), an accepted sibling inserts (a row beside a row), a leaf destination where both derivations agree stays unchanged, and a root-level sibling insert beside the container still works, the contract the container escape behaviors rely on. Blast radius checked by running the full editor suite (1705 browser tests on chromium plus 852 unit tests) and the plugin-table suite, all passing unmodified.

### DIFF
--- a/.changeset/insert-block-sibling-scope.md
+++ b/.changeset/insert-block-sibling-scope.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: validate `insert.block` sibling placements against the parent array's schema
+
+`insert.block` with `placement: 'before'` or `'after'` at a point inside a container validated the block against the destination's own sub-schema view while landing it as a sibling in the parent array. When the two scopes differed, a text block addressed at a table row, for example, the block passed validation and landed as schema-invalid data inside an array that rejects it. Such inserts now no-op, the same outcome as other schema-rejected inserts.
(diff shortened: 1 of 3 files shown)


christianhg:
Closing as part of the open-PR cleanup, with the work fully preserved and verified: rebased onto today's `main`, the bug proven alive first (this PR's row-placement test red on `main`), then the full monorepo browser suite, unit suite, and statics green on the branch. Tracked in Linear for revival; reopening this PR or cherry-picking the branch should cost nothing but trivial drift.

PR #2932, opened 2026-07-07
feat(plugin-table): export the stylesheet as `stylesText`

christianhg:
The driver is the Studio integration of `@portabletext/plugin-table`: the `sanity` package is a styled-components library build with no CSS pipeline, so it cannot `import '@portabletext/plugin-table/ui/styles.css'`, and the integration spike carried a hand-inlined copy of the stylesheet instead. That copy is drift waiting to happen, and dangerously so, because the stylesheet is not just theming: it carries structural rules the chrome depends on (selection hiding, focus rings, the portal font, button resets), all version-locked to the chrome's markup. A host running a stale copy against newer chrome breaks subtly with nothing in the console.

The shape follows the ecosystem's established answer for this exact problem. Styled-chrome libraries ship a CSS file the app imports (BlockNote, ProseMirror, Leaflet), and for hosts that cannot take the file, the precedents are auto-injection with an opt-out (`react-tooltip`, `sonner`) or the file plus the same content as importable JS (`react-toastify`'s `injectStyle`). This PR is the third pattern: `@portabletext/plugin-table/ui` now exports `stylesText`, and Studio interpolates it into a styled-components `createGlobalStyle`, version-locked to whatever plugin version it ships. Auto-injection was considered and rejected: it would bypass every consumer's CSS toolchain to solve one host's build constraint.

The implementation is deliberately mechanism-free: `styles-css-text.ts` does a Vite-style `?raw` import of `styles.css` itself, which vitest resolves natively and the library build resolves through a ten-line rollup shim in `package.config.ts`. The export is byte-identical to the file by construction, so there is no codegen step, no generated module, and no drift test to maintain. The stylesheet file stays the canonical path and the README says so: for hosts with a bundler it participates in the consumer's CSS toolchain (PostCSS transforms such as a `light-dark()` polyfill, content hashing, cascade placement, forkability), which text cannot.

Full plugin suite (120 on chromium) passes; the strict build inlines the stylesheet into `dist/ui/index.js` and declares `stylesText: string` in the d.ts; `check:types`, lint, react-compiler, knip, and format are green.

### DIFF
--- a/.changeset/styles-text.md
+++ b/.changeset/styles-text.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/plugin-table': minor
+---
+
+feat: export the stylesheet as `stylesText` for hosts without a CSS pipeline
+
+`@portabletext/plugin-table/ui` now exports the reference UI's stylesheet as text, generated from `styles.css` so the two cannot diverge. Hosts whose build cannot import a CSS file inject it through whatever styling system they already have, for example a styled-components `createGlobalStyle`. Hosts with a bundler keep importing `@portabletext/plugin-table/ui/styles.css` as before.
--- a/packages/plugin-table/README.md
+++ b/packages/plugin-table/README.md
@@ -179,6 +179,19 @@ The stylesheet ships the structural rules plus a set of
 shared ancestor. The defaults are declared at zero specificity, so any
 declaration of yours wins.
 
+If your build has no CSS pipeline and cannot import the stylesheet file,
+`stylesText` is the same content as a string, generated from the
+stylesheet so the two cannot diverge. Inject it through whatever styling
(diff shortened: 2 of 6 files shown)


christianhg:
Closing unmerged: the investigation this triggered invalidated its own premise. The `sanity` package now has a CSS pipeline (pkg-utils with `vanillaExtract: true`, `Studio.tsx` imports `./styles.css`, the build emits `sanity/bundle.css`), and the repo's bundler tests explicitly pin that third-party dependency CSS imports (`import 'awesome-datepicker/styles.css'") survive into the built output. So the Studio integration uses the plain stylesheet import, and `stylesText`'s only named consumer is gone. If a genuinely pipeline-less host shows up, this branch is a cherry-pick away.

PR #2958, opened 2026-07-09
fix: expand marks the schema cannot resolve when typing at span edges

christianhg:
Typing at the trailing edge of a span whose mark the schema cannot resolve started a new unmarked span, the text escaped the mark the way it escapes an annotation. For a decorator that was merely removed from the schema (or one a collaborator has and you don't), that's the wrong instinct: since 7.10.4 such marks are preserved as possible decorators, so editing should give them decorator semantics too, and decorators expand.

The cause is the same closed-world inference #2954 removed from normalization, still alive in the typing path: the caret branch of `getMarkState` and `getActiveAnnotationsMarks` classified every mark that isn't a schema decorator as an annotation. A mark is now an annotation only when it resolves to one of the block's `markDefs`, the line the expanded-selection branch of the same selector, the render classification, and normalization already draw. The `insert.text` behavior carries unresolvable marks onto inserted text, excluding declared decorators by name rather than active state so a toggled-off decorator doesn't ride along (the existing `Toggling bold inside italic as you write` scenario pins exactly that, and caught exactly that during development).

Four new scenarios in `tests/schema-less-marks.test.tsx`: the decorator-expansion and annotation-escape baselines, typing inside an unresolvable mark, and the trailing-edge expansion (red before the fix). Two `getMarkState` unit fixtures gained the `markDefs` they always meant to have: the old classifier never read markDefs, so their annotation marks referenced none, which the new classifier correctly treats as unresolvable. Full suites: 1718 browser, 852 unit.

### DIFF
--- a/.changeset/unknown-mark-typing-semantics.md
+++ b/.changeset/unknown-mark-typing-semantics.md
@@ -1,0 +1,10 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: expand marks the schema cannot resolve when typing at span edges
+
+Typing at the edge of a span whose mark the schema cannot resolve now
+extends the span, the way decorators behave. Previously such marks
+were treated as annotations and escaped, so the typed text started a
+new unmarked span next to the marked one.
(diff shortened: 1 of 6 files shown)


christianhg:
Parking this for now, it works and is fully pinned, but it's not a priority. The branch stays for a future resume.

PR #2974, opened 2026-07-15
fix: apply remote patches addressing marks and markDefs elements as data patches

ryanbonial:
## Description

Remote patches produced by diffing tools (e.g. `@sanity/diff-patch` in `@portabletext/plugin-sdk-value`) can address elements of sidecar arrays: `span.marks[0]`, `span.marks[-1]`, `block.markDefs[_key=="..."]`. These paths end in a keyed or numeric segment just like structural node paths, so `applyOperation` routed them through structural child insertion/removal against the owning node's `children`.

The consequences in a two-editor session, whenever a collaborator toggles a decorator or annotation:

- `unset marks[0]` / `unset markDefs[_key=="..."]` threw `Cannot apply an "unset" (node removal) operation ... because the node was not found.`, killing the consumer's sync actor. The editor kept accepting keystrokes but stopped receiving remote changes until reload.
- `insert after marks[-1]` silently wrote a bogus `children: ['strong']` array onto the span. The corrupted span was then pushed back to the datastore and persisted.
- `set marks[0]` and `diffMatchPatch` on `marks[0]` were silently ignored, so the editors drifted apart.

The fix teaches the engine to detect that such a path does not target the owning node's structural child array (`targetsStructuralChildren`) and to apply the operation as a plain data patch on the root block (`applyOnRootBlock`, via `applyAll` from `@portabletext/patches`), matching what the datastore computed. This is the same fallback pattern the `set` and property-`unset` handlers already use for markDefs property paths. An alternative of rewriting these patches in `plugin-sdk-value` was rejected: it would only cover SDK apps and leave the engine bug reachable from every other remote patch source.

## What to review

- `packages/editor/src/engine/core/apply-operation.ts`: the routing check and the data-patch fallback, including the `set`-based inverse computed for local sidecar unsets (remote patches skip inverses).
- `packages/editor/src/internal-utils/applyPatch.ts`: the `diffMatchPatch` fallback for non-`text` string targets.
- Whether `getChildFieldName` is the right structural-field oracle for the check.

## Testing

- `packages/editor/tests/event.patches.sidecar-arrays.test.tsx`: 6 regression tests using the exact patch shapes `diffValue` emits for decorator/annotation toggles. All 6 fail on `main` (three throw, three silently corrupt or no-op).
- `packages/plugin-sdk-value/src/plugin.value-sync.browser.test.tsx`: 2 end-to-end tests driving `ValueSyncPlugin` through remote decorator and annotation toggles, covering the span split/merge + `markDefs` flow.
- Full existing suites pass: editor chromium browser tests (1721), editor unit tests (852), plugin-sdk-value suites, typecheck, lint.

### DIFF
--- a/.changeset/sidecar-array-remote-patches.md
+++ b/.changeset/sidecar-array-remote-patches.md
@@ -1,0 +1,24 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: apply remote patches addressing `marks` and `markDefs` elements as data patches
+
+Remote patches produced by diffing tools (e.g. `@sanity/diff-patch` in
+`@portabletext/plugin-sdk-value`) can address elements of sidecar
+arrays: `span.marks[0]`, `span.marks[-1]`, `block.markDefs[_key==...]`.
+These paths end in a keyed or numeric segment just like structural node
+paths, so the engine routed them through structural child
+insertion/removal. Removals threw `Cannot apply an "unset" (node
+removal) operation ... because the node was not found.` and killed the
+consumer's sync, while inserts silently wrote a bogus `children` array
+onto the span, corrupting the document once the value was pushed back
+to the datastore. In practice this broke concurrent editing whenever a
+collaborator toggled a decorator or annotation.
+
+The engine now detects that such a path does not target the owning
+node's structural child array and applies the operation as a plain data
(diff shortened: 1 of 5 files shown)


christianhg (approved):
This looks great, thank you @ryanbonial!

PR #2977, opened 2026-07-16
feat(plugin-sdk-value): sync through operational patches instead of whole-value diffs

ryanbonial:
### Description

The plugin currently syncs by diffing whole-value snapshots in both directions, which makes concurrent edits to the same field last-writer-wins at best. `@sanity/sdk` 2.17.0 added an operational patch channel (`remote-patches` document events on the read side, `editDocument(..., {preserveOperations: true})` on the write side, see sanity-io/sdk#1046), and this PR makes the plugin use it.

- Incoming: `SDKValuePlugin` subscribes to `remote-patches` events, scopes document-rooted patches to the configured field, and applies them directly to the editor in every sync state. The editor merges them with in-flight local changes.
- Outgoing: the editor's own mutation patches are converted to Sanity patch operations and dispatched with `preserveOperations`, so keyed operations reach the server intact.
- Fallbacks: whole-value diff sync remains for sync events, patches that cannot be scoped or converted, and older SDK versions. Against `@sanity/sdk-react` < 2.17.0 the plugin behaves exactly as before.

`ValueSyncPlugin` gains optional `onRemotePatches`/`pushPatches` config for custom stores.

### Repair sync after best-effort patch application

Applying remote patches is best-effort: keyed operations produced by another client against a different view of the document (for example both clients splitting the same span to format overlapping ranges) can fail to resolve and are skipped, leaving the editor diverged from the stored document. The second commit closes that gap:

- Every remote patch application is now followed by a whole-value repair sync, immediately when no local edits are in flight or after the next mutation flush when they are. If the diff-based repair cannot converge, the plugin escalates to the full value sync machinery.
- The repair diff no longer emits operations addressing items inside sidecar arrays (`markDefs`, `marks`), which the engine cannot resolve. Those are coalesced into whole-property `set` operations taken from the target value.

### Testing

- Unit tests for `stringifyPatchPath`, `convertPatchesToSanity`, `scopeRemotePatches`, and the new `findSidecarArrayPath`/`toEngineSafePatches` helpers.
- Browser tests covering remote patches applying operationally, applying mid-typing without losing local changes, local edits pushing patches with the store converging purely through `applyAll`, the whole-value fallback when patch pushing throws, and repair convergence when remote patches cannot apply.
- A new two-client browser rig (`plugin.two-client.browser.test.tsx`) connects two real editors through a mock server that applies patches best-effort and rebases each client's pending local changes, mirroring SDK optimistic behavior. It verifies convergence for concurrent typing in different blocks, concurrent same-range bolding, overlapping bold and link over the same range, and formatting conflicts during concurrent typing.
- Verified end-to-end against a real dataset: two independent `SanityInstance` clients editing the same field through the Content Lake listener on `@sanity/sdk-react` 2.17.0, with concurrent typing and same-range formatting converging on both sides.

### DIFF
--- a/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
+++ b/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
@@ -261,6 +261,118 @@ function segmentsEqual(a: PathSegment, b: PathSegment): boolean {
   return a._key === b._key
 }
 
+function pathsEqual(a: Path, b: Path): boolean {
+  return (
+    a.length === b.length &&
+    a.every((segment, index) => segmentsEqual(segment, b[index]))
+  )
+}
+
+/**
+ * The editor engine can only resolve keyed/indexed segments against the
+ * root block array and (possibly nested) `children` arrays. Operations
+ * that address items of any other array, e.g. a block's `markDefs` or a
+ * span's `marks`, misapply. When a patch path enters such a sidecar
+ * array, this returns the path of the array itself so callers can fall
+ * back to replacing the whole property; returns `null` for paths the
+ * engine can apply.
+ *
+ * @internal
+ */
+export function findSidecarArrayPath(path: Path): Path | null {
+  // a keyed/numeric segment is only resolvable first (root block array)
+  // or directly after a `children` property
+  let expectNode = true
+  for (let index = 0; index < path.length; index++) {
+    const segment = path[index]
+    if (typeof segment === 'string') {
+      expectNode = segment === 'children'
+    } else {
+      if (!expectNode) {
+        return path.slice(0, index)
+      }
+      expectNode = false
+    }
+  }
+  return null
+}
+
+function getValueAtPath(value: JSONValue, path: Path): JSONValue | undefined {
+  let current: JSONValue | undefined = value
+  for (const segment of path) {
+    if (current === null || current === undefined) {
+      return undefined
+    }
+    if (typeof segment === 'number') {
+      if (!Array.isArray(current)) {
+        return undefined
+      }
+      current = current[segment]
+    } else if (typeof segment === 'string') {
+      if (typeof current !== 'object' || Array.isArray(current)) {
+        return undefined
+      }
+      current = (current as {[key: string]: JSONValue})[segment]
+    } else if (Array.isArray(segment)) {
+      // index tuples address ranges, not single values
+      return undefined
+    } else {
+      if (!Array.isArray(current)) {
+        return undefined
+      }
+      current = current.find(
+        (item) =>
+          typeof item === 'object' &&
+          item !== null &&
+          !Array.isArray(item) &&
+          (item as {_key?: unknown})._key === segment._key,
+      )
+    }
+  }
+  return current
+}
+
+/**
+ * Converts a target-value diff into patches the editor engine can apply.
+ * Patches addressing items inside sidecar arrays are coalesced into whole
+ * `set`s (or `unset`s) of the owning property, taken from the target
+ * value.
+ *
+ * @internal
+ */
+export function toEngineSafePatches(
+  patches: PtePatch[],
+  targetValue: PortableTextBlock[],
+): PtePatch[] {
+  const safe: PtePatch[] = []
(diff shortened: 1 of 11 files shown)


ryanbonial:
Update: root cause of the remaining harness crashes found, and it was in the SDK, not this plugin.

The `Unexpected node type` crashes in `bold-overlap-range` / `link-overlap-range` came from `@sanity/sdk`'s local patch applier fabricating array items for unresolvable keyed `set` paths (e.g. `children[_key=="X"].marks` after a concurrent split removed span X). Content Lake silently no-ops those, so the server stayed valid while every receiving client's local snapshot grew phantom span-less `{marks: []}` children; this plugin's repair sync then faithfully converged the editor to the corrupted snapshot and the render layer threw. Fixed SDK-side in sanity-io/sdk#1059 with crash forensics and harness runs in the PR description (0 crashes across the full 13-scenario matrix after the fix, `bold-overlap-range` and `link-overlap-range` now score SAFE).

The commit just pushed here (6e26dce) is test-only: the vendored mock-server applier in the collision matrix mirrored the SDK's fabrication bug, making the mock more forgiving than the real backend. It now matches Content Lake (unresolvable keyed/indexed `set` segments are no-ops, verified against gradient), and the full collision matrix still passes against the corrected semantics.

For harness verification of this PR, pair it with an SDK build that includes sanity-io/sdk#1059.

packages/plugin-sdk-value/src/plugin.sdk-value.tsx
```diff
@@ -152,6 +157,404 @@ export function convertPatches(patches: SanityPatchOperations[]): PtePatch[] {
...
+
+/**
+ * Converts a target-value diff into patches the editor engine can apply.
+ * Patches addressing items inside sidecar arrays are coalesced into whole
+ * `set`s (or `unset`s) of the owning property, taken from the target
+ * value.
+ *
+ * @internal
+ */
+export function toEngineSafePatches(
```
christianhg, line 349: Should we delete this? It's probably not needed after #2974, right?

packages/plugin-sdk-value/src/plugin.sdk-value.tsx
```diff
@@ -152,6 +157,404 @@ export function convertPatches(patches: SanityPatchOperations[]): PtePatch[] {
...
+ * operations against the store (server truth): new definitions insert,
+ * changed definitions set by key, and removed definitions unset by key,
+ * except definitions the store's own spans still reference (a diverged
+ * client's normalizer prunes those spuriously; a later, converged flush
+ * removes them for real). Item-keyed operations merge at the server
+ * instead of overwriting.
+ *
+ * @internal
+ */
+export function toMergeableMarkDefsPatches(
```
christianhg, line 397: Feels like one more thing to fix in the editor engine. I'll make a note of that.

christianhg (approved):
Nice!

packages/plugin-sdk-value/src/plugin.sdk-value.tsx
ryanbonial: I tested exactly this before answering 🧑‍🔬: with toEngineSafePatches turned into a pass-through on this branch (which has #2974 in its history), 15 browser tests fail, specifically link-same-range and link-overlap-range in both delivery orders in the collision matrix plus the two-client overlapping link+bold test. Restoring it, all 72 pass.

They two of them address different layers. #2974 makes the engine mechanically able to apply "sidecar" item operations (routing them as data patches instead of misrouting into children). toEngineSafePatches guards against semantic misapplication under divergence: these patches are produced by diffing against the store's value, so an item-level op like insert after `markDefs[-1]` is only correct relative to the store's version of the array. Applied to an editor whose local `markDefs` differ mid-conflict, it appends against the wrong state (duplicate/stale defs), even though the engine applies it fine. Coalescing to a whole-property set from the target value converges regardless of local state, which is what the repair path needs. Agree it's another candidate for eventually moving into the engine, alongside your `toMergeableMarkDefsPatches` note below.

christianhg: Thank you, that makes sense!

PR #2982, opened 2026-07-20
fix: emit item-keyed `markDefs` patches instead of whole-array sets

christianhg:
**Stacks on #2983 (undo-step boundaries) and #2984 (plugin guard).** The first two commits here are those PRs verbatim and disappear on rebase once they merge; review only the last commit in this PR. Both are extracted because they merge independently: #2983 fixes latent undo bugs reachable on `main` today, and #2984 must release ahead of this PR's editor change (the plugin's published `^7.x` peer range accepts the new emission, so a consumer upgrading the editor without the plugin would otherwise run the orphan race unguarded).

Two clients annotating the same block concurrently silently destroy each other's work: client A adds a comment annotation, client B links a word in the same block, and whoever flushes last wins. The loser's definition vanishes from the stored document while both clients' span `marks` references survive, leaving marks pointing at definitions that no longer exist. The concrete driver is the operational patch channel in #2977, which compensates for this outgoing shape channel-side (`toMergeableMarkDefsPatches`); fixing the emission in the engine pays for every multi-writer consumer at once, including two Studio users on the same block.

The granularity was lost at the operation level: every `markDefs` mutation funneled through `setNodeProperties`, which applies one `set` operation carrying the whole array, and patch generation echoed it verbatim. An earlier iteration of this PR fixed the symptom in the operation-to-patch layer by diffing the pre-apply value, but that reconstructs intent the operation sites had precisely and then erased, a translation layer compensating for the layer above it. The shipped shape moves the fix to the sites: `annotation.add` and the block-merge path in `insert.block` apply keyed `insert` operations, and the normalizer's unused-definition prune applies keyed `unset`s. The engine's sidecar routing (shipped in #2974 for remote patches) applies these locally unchanged, and now also computes item-keyed inverses so undo emits mergeable patches too instead of restoring whole arrays. Patch generation shrinks to mechanical representation changes, with the invariant pinned in a header comment: never diff sibling state to reconstruct intent here; if a patch needs more than the operation expresses, the operation site is emitting the wrong operation. One translation remains by necessity: normalization's ensure-`markDefs`-exists `set([], ...)` becomes `setIfMissing`, because no operation vocabulary carries ensure-exists intent, and a whole-array set of `[]` would clobber concurrent writers on documents that lack the property.

Inserts anchor on index 0 rather than on an existing definition's key: a keyed anchor diffed against this engine's state may no longer exist at the server by application time (the same divergence lesson as the repair-path discussion on #2977), while index 0 always resolves and `markDefs` order carries no meaning. New definitions therefore prepend instead of append, a value-visible order change pinned by re-pinned tests in `overlapping-annotations` and `unique-sibling-keys`. Explicit whole-array writes through `block.set` still emit a whole-array `set` (whole-property intent from a public API, and duplicate-key arrays, which the dedupe rule exists for, are inexpressible item-level).

Removing the always-applied whole-array set unmasked two latent history bugs, fixed in their own commit. The `isNoOp` early-return in the history subscriber advanced `previousUndoStepId`, so a no-op leading a new undo step consumed the step boundary and a single undo reverted more than one step's worth of changes. And a step's pre-state selection was captured from the first history-affecting operation, by which time operation implementations may have parked the selection on transient nodes their own step removes again; the engine now snapshots `undoStepSelection` when the undo step ID is minted. Both were masked on `main` because the redundant `markDefs` set reached the history subscriber first, creating the boundary and capturing the selection while it was still resolvable. The `event.history.undo`/`event.history.redo` suites pin the corrected behavior.

The third commit extends `plugin-sdk-value`'s referenced-keys guard to the new patch shape. The editor's keyed prune unsets passed through `toMergeableMarkDefsPatches` untouched, bypassing the guard that previously protected decomposed whole-array sets, and a diverged client could prune a definition the store's spans still reference, exactly the orphaned-mark failure the collision-matrix link scenarios caught (`orphan mark (no markDef)`). Keyed `markDefs` unsets now go through the same store-side check; a dropped prune leaves an unused definition behind at worst, which a later converged session's normalization removes. The full collision matrix and the two-client rig pass on the new emission.

The emit-to-apply round trip is pinned by the browser scenario "Emitted patches converge another editor on the same value" in `event.patch.markdefs.test.tsx`, and the ensure-exists translation by the `setNodePatch` unit matrix. The 12 existing tests pinning whole-array shapes are re-pinned; the remaining editor suites (857 unit, 1733 browser) and all `plugin-sdk-value` suites (192 across chromium, firefox and webkit) pass.

The one deliberately unsolved residual: the prune-vs-concurrent-reference race is inherent to `marks` being string references into a separate array, and fully closing it engine-side needs store knowledge the engine lacks. The channel-side guard covers the SDK path; deferred pruning as a general mechanism is tracked separately as a follow-up.

### DIFF
--- a/.changeset/guard-keyed-markdefs-unsets.md
+++ b/.changeset/guard-keyed-markdefs-unsets.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/plugin-sdk-value': patch
+---
+
+fix: guard outgoing keyed `markDefs` unsets against store-referenced definitions
+
+An upcoming `@portabletext/editor` release prunes unused annotation definitions as item-keyed `unset` patches instead of whole-array sets. A client that has diverged from the store can prune a definition the store's spans still reference, which would orphan another client's annotation. Outgoing keyed `markDefs` unsets are now dropped while the store's spans reference the definition, mirroring the existing guard on decomposed whole-array sets; at worst an unused definition lingers until a later converged session prunes it. Against current editor releases the guard is inert.
--- a/.changeset/markdefs-item-patches.md
+++ b/.changeset/markdefs-item-patches.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: emit item-keyed `markDefs` patches instead of whole-array sets
+
+Changing a block's annotations previously emitted a `set` of the whole `markDefs` array. When two clients annotated the same block concurrently, the last writer's array overwrote the other's at the server while both clients' span `marks` references survived, leaving annotations without definitions. Added definitions are now emitted as a `setIfMissing` plus an item `insert`, removed definitions as keyed `unset`s, and undoing those edits emits the item-keyed counterparts. Item-keyed operations merge at the server instead of overwriting.
+
+New definitions are prepended to `markDefs` instead of appended, so code that reads the array positionally sees new definitions first. Consumers observing `patch` or `mutation` events see the finer-grained shapes; explicit whole-array writes through `block.set` still emit a whole-array `set`.
--- a/.changeset/undo-step-boundaries.md
+++ b/.changeset/undo-step-boundaries.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: keep leading no-op operations from consuming undo-step boundaries
+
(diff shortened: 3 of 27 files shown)


christianhg:
Closing as part of the open-PR cleanup. The stack it headed (#2983, #2984) is parked in tickets with both extractions verified green as of today; this branch (`fix-markdefs-item-patches`) preserves the editor-side emission change as its last commit, and rebasing it onto the two merged extractions was always the plan. Tracked in Linear.

PR #2983, opened 2026-07-20
fix: keep leading no-op operations from consuming undo-step boundaries

christianhg:
A single undo can revert more than the most recent edit. The history subscriber's `isNoOp` early-return (operations without an inverse, zero-length text operations) advanced `previousUndoStepId` before bailing, so a no-op leading a new undo step consumed the step boundary: the next history-affecting operation saw `currentUndoStepId === previousUndoStepId` and merged into the previous step. One undo then reverted two edits' worth of changes. `previousUndoStepId` now only advances for operations that reach `createUndoSteps`, since its job is to track the last operation that affected history.

This is reachable on `main` through public API: `child.set` replacing a span's text leads with a `remove.text` of the old text, which is zero-length when the span was empty. The new scenario "Undoing a `child.set` that replaces an empty span's text" pins it: two edits through `execute` actions, one undo, only the second edit reverts. On `main` the same test fails with both edits reverted. The producer isn't unique to `child.set`: `applySplitNode` emits an unguarded zero-length `remove.text` when splitting a span at its end, which the annotation and block-merge flows reach today, masked only because an unconditionally applied whole-array `markDefs` set happens to hit the history subscriber first. #2982 removes that masking operation, which is how both defects surfaced; it stacks on this.

The same masking hid a second defect: a step's pre-state selection was captured from the first history-affecting operation's `beforeSelection`, but operation implementations can park the selection on transient nodes (created and removed within the same step) before that operation applies. Undo then restored a selection that no longer resolves and the cursor disappeared. The engine now snapshots its selection as `undoStepSelection` at the moment `editor.undoStepId` is minted, and the history subscriber uses that snapshot as the pre-state selection whenever an operation opens a new step. This half is exercised without the mask by the custom-event undo scenarios in #2982; on `main` no unmasked flow reaches it, so its pin lives there.

Net behavior unchanged for flows whose steps begin with a history-affecting operation; the full unit and browser suites pass without modification.

### DIFF
--- a/.changeset/undo-step-boundaries.md
+++ b/.changeset/undo-step-boundaries.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: keep leading no-op operations from consuming undo-step boundaries
+
+A single undo could revert more than the most recent edit: when an editing flow's first operation was one that doesn't affect history (a zero-length text change, or an operation without an inverse), the following operations merged into the previous undo step. Undo now reverts exactly one step.
+
+Undo also restores the selection from before the step as it was when the edit began, instead of a mid-flow selection that may reference nodes the undo itself removes (which dropped the cursor entirely).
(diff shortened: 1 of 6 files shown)


christianhg:
Closing as part of the open-PR cleanup, with the work verified and preserved: rebased onto today's `main`, undo suite, unit suite, and CI green on `fix-undo-step-boundaries`. What's pending is not mechanics but a line-by-line review of the step-boundary semantics (whether any path now splits steps that previously merged on purpose). Tracked in Linear; reviving this is the first step of reviving the item-keyed `markDefs` line.

PR #2984, opened 2026-07-20
fix(plugin-sdk-value): guard keyed `markDefs` unsets against store-referenced definitions

christianhg:
Preparatory extraction from #2982, which changes the editor to emit normalization's unused-annotation-definition prunes as item-keyed `unset` patches instead of whole-array `markDefs` sets. Under that emission, a client that has diverged from the store (its span `marks` overwritten by another client's flush, its own definition now locally unused) prunes a definition the store's spans still reference. `toMergeableMarkDefsPatches` only decomposes whole-array sets, so the keyed unset passed through without the referenced-keys guard and orphaned the other client's annotation at the server; the collision-matrix link scenarios in this package caught exactly that (`orphan mark (no markDef)`) when run against #2982's emission.

Keyed `markDefs` unsets now route through the same store-side referenced-keys check as decomposed whole-array sets, and are dropped while the store's spans reference the definition. A dropped prune leaves an unused definition behind at worst, which normalization in a later converged session removes for real, strictly better than a dangling mark.

This lands and releases ahead of #2982 because of the published peer range: the plugin's `^7.x` peer accepts an editor with the new emission, so a consumer upgrading `@portabletext/editor` without upgrading this plugin would run the orphan race with no guard. Shipping the guard first closes that window. Against current editor releases the new branch is inert (keyed `markDefs` unsets don't occur), which is also why all suites here pass unchanged on `main`'s emission; the collision-matrix runs that exercise the guard live in #2982, which stacks on this.

### DIFF
--- a/.changeset/guard-keyed-markdefs-unsets.md
+++ b/.changeset/guard-keyed-markdefs-unsets.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/plugin-sdk-value': patch
+---
+
+fix: guard outgoing keyed `markDefs` unsets against store-referenced definitions
+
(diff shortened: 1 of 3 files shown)


christianhg:
Closing as part of the open-PR cleanup, with the work verified and preserved: rebased onto today's `main`, 216/216 tests, statics, and CI green on `fix-guard-keyed-markdefs-unsets`. The guard is dormant until the editor's keyed `markDefs` emission ships, and that line is parked in tickets; sequencing note for revival: this plugin change must release ahead of the editor emission change.

PR #2986, opened 2026-07-21
fix(plugin-sdk-value): stop dropping mutation flushes and stomping in-flight typing

ryanbonial:
## The bug

A single user typing normally, no collaboration involved, could see their text reordered, duplicated, or resurrected, most visibly after deleting a few characters and retyping them. The editor and the server converge on the mangled result, so it looks like this:

```
typed:  "...briefing exclusive sanity context briefing reporting briefing"
result: "...briefing exclusive e sanity context text briefing reing porting briefingfing"
```

Reported in the field against `@sanity/sdk-react@2.18.0` + `@portabletext/plugin-sdk-value@7.1.0` (video showed text going out of order while one person typed). Reproduced locally with a Playwright driver that types through the contenteditable with human keystroke timing and delete-and-retype bursts: **6 of 6 trials mangled** on those versions. The old whole-value plugin (7.0.34) mangled 1 of 4 trials with the same gesture, so this predates the patch channel but the patch channel made it consistent.

## Root cause

Two sync-machine bugs compounding (instrumented traces in the investigation showed both):

1. **Dropped mutation flushes.** The editor emits `mutation` events in bursts, one per pending operation, so a quick backspace run produces several flush events back to back. The machine only handled `mutation flushed` in `local write` and `pending sync`. A flush arriving in `pushing to remote` or `idle` (the very next event in a burst) was silently dropped: its patches never reached the store, and the store diverged from the editor permanently. The patch channel is more sensitive to this than the old plugin, because a whole-value push self-heals drift on the next flush while incremental patches accumulate it.

2. **Mid-typing whole-value repair.** The machine ran its editor-versus-store repair diff immediately after flushes (`defer then apply sync`) and on remote patch arrivals, while the user was still typing. Diffing against a store that lags the user's keystrokes (or has drifted per bug 1) injects the difference back into the editor as fake remote patches: it resurrects text the user just deleted and re-asserts stale word suffixes ("deadline" + "dline", "editors" + "s"). The next flush pushes the corruption to the server, and both sides converge on the mangled text.

## The fix

- `mutation flushed` now pushes in **every** state, so no flush is ever dropped.
- The whole-value repair only runs when the editor is quiescent: on a store change while `idle`, plus a one-shot repair after a short idle delay (500ms) to heal residual divergence from concurrent editing when no further store event arrives. All mid-typing repair triggers are removed.
- An opt-in sync diagnostics channel stays behind `globalThis.__PTE_SYNC_DEBUG` (zero cost when off); it is what made the traces above possible.

## Tests

- New browser regression test (`plugin.value-sync.browser.test.tsx`): drives real keystrokes via `userEvent` (each key its own operation, which is what produces the flush bursts), performs the delete-and-retype gesture across flush windows without waiting for convergence, and requires the editor, the store, and a keystroke-perfect expected text to all agree. Gate-checked: it fails on the previous machine in chromium, firefox, and webkit with the exact duplication signature from the field report, and passes with the fix.
- Full package suite green: 199 tests across all three browsers, types, lint, build.

## Verification beyond the unit level

Against real Content Lake with the external two-client harness (sanity-labs/pte-sdk-concurrent-repro):

- Solo-typing driver: 6/6 SAFE standalone and 6/6 SAFE embedded in a Studio tool (was 6/6 mangled).
- Full 17-scenario concurrent matrix: all typing/delete scenarios SAFE, 0 crashes, 0 server-document corruption; overlapping-formatting soft-loss rates unchanged. The repair-timing change does not regress two-client convergence.
- Verified end to end in the SDK kitchensink app against this build.

Independent of #2985 (different files); both can land in either order.

### DIFF
--- a/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
+++ b/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
@@ -555,6 +555,35 @@ export function scopeRemotePatches(
   return scoped
 }
 
+// Sync diagnostics, disabled unless `globalThis.__PTE_SYNC_DEBUG = true`
+// is set before load. Callers pass a thunk so detail computation is free
+// when the flag is off.
+function syncDebugEnabled(): boolean {
+  return (globalThis as {__PTE_SYNC_DEBUG?: boolean}).__PTE_SYNC_DEBUG === true
+}
+
+function syncDebug(kind: string, detail: () => Record<string, unknown>) {
+  if (syncDebugEnabled()) {
+    // biome-ignore lint/suspicious/noConsole: opt-in diagnostics channel
+    console.debug(`[pte-sync] ${kind} ${JSON.stringify(detail())}`)
+  }
+}
+
+function debugTextOf(value: unknown): string {
+  if (!Array.isArray(value)) {
+    return ''
+  }
+  return value
+    .map((block) =>
+      Array.isArray((block as {children?: unknown}).children)
+        ? ((block as {children: Array<{text?: string}>}).children ?? [])
+            .map((child) => child.text ?? '')
+            .join('')
+        : '',
+    )
+    .join('\n')
+}
+
 function applySync({
   editor,
   getRemoteValue,
@@ -569,6 +598,13 @@ function applySync({
   }
 
   const snapshot = editor.getSnapshot().context.value
+  if (syncDebugEnabled()) {
+    const editorText = debugTextOf(snapshot)
+    const remoteText = debugTextOf(remoteValue)
+    if (editorText !== remoteText) {
+      syncDebug('applySync:texts-differ', () => ({editorText, remoteText}))
+    }
+  }
 
   let patches: PtePatch[]
   try {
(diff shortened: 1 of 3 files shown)


packages/plugin-sdk-value/src/plugin.sdk-value.tsx
```diff
@@ -555,6 +555,35 @@ export function scopeRemotePatches(
   return scoped
 }
 
+// Sync diagnostics, disabled unless `globalThis.__PTE_SYNC_DEBUG = true`
+// is set before load. Callers pass a thunk so detail computation is free
+// when the flag is off.
+function syncDebugEnabled(): boolean {
```
christianhg, line 561: Would probably be nice to adopt the `debug` library in this package as well. We could give logs a namespace like `pte:plugin-sdk-value:*` to interleave them nicely with existing editor sync logs like `pte:mutation` and `pte:sync:value`.

https://github.com/portabletext/editor/blob/cc75b6f3ab341014f618fd7abf4743750bf65ddb/packages/editor/src/editor/sync-machine.ts#L177

Something to follow up on, not a blocker.

christianhg (approved):
Looks GOOD!

PR #3000, opened 2026-07-22
fix(plugin-sdk-value): drop remote normalize-merge mutation echoes

ryanbonial:
## Description

Receiving editors were pushing their own normalize fallout after applying a collaborator's formatting (adjacent same-mark span merges). Those competing merges truncated or duplicated text on the originator's block under concurrent edit (Axios format-while-peer-types).

Studio FormBuilder does not show this class for the same gesture; the SDK plugin path does because it eagerly pushes every mutation flush. This change arms a short gate after remote apply and drops a flush only when its patches look like a normalize merge (`unset` of a keyed `children` member plus text `diffMatchPatch`/`set`). Ordinary typing flushes are unaffected.

Supersedes the approach in #2999 (core `withoutPatching` nesting), which broke self-solving tests. Prefer fixing at the plugin push boundary.

**Not in scope:** paste-count loss, same-block typing scramble, overlapping-format SILENT-LOSS. Those remain a separate roadmap; Phase 0 research will decide whether follow-ups branch from this PR or from `main`.

## What to review

- `packages/plugin-sdk-value/src/plugin.sdk-value.tsx`: `remoteApplyEchoGate` / `isNormalizeMergeEcho` / drop in `listenToEditor`
- `packages/plugin-sdk-value/src/plugin.value-sync.browser.test.tsx`: echo not pushed; unpushed typing still pushed after remote normalize

## Testing

- Plugin browser suite: 24/24 including new regression tests
- External harness (`pte-sdk-concurrent-repro` on local plugin build):
  - `duo-typing.mjs format`: block A text integrity restored vs released (released still mangled)
  - `duo-typing.mjs blocks` / `solo-typing.mjs`: green after shape-based gate (value-equality gate had caused blocks scramble; fixed before this PR)
  - Full matrix still has known reds (paste counts, same-block, overlapping SILENT-LOSS); not claimed fixed here

## Notes for release

Customer-facing: concurrent bold/format while another user types in another block should no longer scramble the formatted block's text via competing normalize merges when using `@portabletext/plugin-sdk-value`.

### DIFF
--- a/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
+++ b/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
@@ -600,6 +600,108 @@ const pendingRepairs = new WeakMap<Editor, PendingRepair>()
  */
 const unflushedEdits = new WeakMap<Editor, boolean>()
 
+/**
+ * After applying remote patches, the editor normalizes (e.g. merges adjacent
+ * same-mark spans) and those merge ops are later flushed as a `mutation`.
+ * Pushing that flush makes every receiver fight the originator with competing
+ * merges. We arm a short gate after each remote apply and drop a flush only
+ * when its patches look like that merge (unset a keyed child + text dmp/set).
+ *
+ * Value-equality alone is not enough: unpushed local typing is already in the
+ * post-apply editor snapshot, so dropping "value unchanged" flushes ate
+ * keystrokes (blocks-mode scramble). Studio's BufferedDocument absorbs the
+ * same fallout by rebasing; we drop it at the push boundary instead.
+ */
+type RemoteApplyEchoGate = {
+  /** Wall-clock deadline; unused gates must not suppress forever. */
+  expiresAt: number
+}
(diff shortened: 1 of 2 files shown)


ryanbonial:
## Confirmation vs #3001 / `@portabletext/editor@7.10.11`

Harness rebaseline on **released** `@portabletext/editor@7.10.11` + `@portabletext/plugin-sdk-value@7.1.3` (no push-boundary echo gate from this PR):

| Cell | Result |
| --- | --- |
| `duo-typing.mjs format` | **4/4 SAFE** (first run); **confirming second run** |
| Completed trials MANGLED | **0** |
| Plugin contains `remoteApplyEchoGate` | **No** (7.1.3) |

Agreed with the #3001 rationale: skipping cosmetic normalize on remote apply removes the echo at the source, so this plugin push gate has nothing left to drop. Keeping a local-only merge would still fork span structure from the store.

**Recommendation:** close this draft as superseded by #3001. Residual roadmap work (paste counts, same-block scramble, overlapping-format SILENT-LOSS) is separate and does not need this branch.

PR #3009, opened 2026-07-22
fix(editor): relocate mid-word remote span DMP inserts past local growth

ryanbonial:
## Description

Same-span End typing can leave one client ahead of a peer's short-base span `diffMatchPatch`. Applying that DMP with `allowExceedingIndices` inserts at the shared prefix and glues words together (`deadlineexclusive`). Plugin-level LCP merges could not fix this reliably because the first break is in the editor apply path.

This change detects prefix-insert DMPs (including trailing EQUAL context) whose naive apply would glue non-space characters, and appends the peer insert at End instead. Spaced mid-prefix inserts on an unchanged base are unchanged.

This is necessary but not sufficient for the same-block harness: editor-only runs still mangle via interleaved / multi-step patches. Seeking direction on a fuller transform strategy vs keeping this as a surgical guard.

## What to review

- `resolve-remote-span-text-dmp.ts`: when we relocate vs fall through
- `applyPatch.ts`: remote-only wiring
- Whether mid-word-only relocation is the right bar, or concurrent End should always append

## Testing

- Unit: `resolve-remote-span-text-dmp.test.ts`
- Browser: `event.patches` scenario "Concurrent End append from a short peer base"
- Harness (`pte-sdk-concurrent-repro`, local editor lib): same-block still mostly MANGLED; this PR does not claim harness green

## Notes for release

N/A – Internal only until behavior is signed off.

### DIFF
--- a/packages/editor/src/internal-utils/applyPatch.ts
+++ b/packages/editor/src/internal-utils/applyPatch.ts
@@ -22,6 +22,7 @@ import {getNode} from '../traversal/get-node'
 import type {PortableTextEditorEngine} from '../types/editor-engine'
 import {applyDeselect} from './apply-selection'
 import {getValue} from './get-value'
+import {resolveRemoteSpanTextDmp} from './resolve-remote-span-text-dmp'
 import {isEqualToEmptyEditor} from './values'
 
 /**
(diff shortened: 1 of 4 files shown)


ryanbonial:
Closing since this changes the contract of DMP without increasing test harness success

PR #3015, opened 2026-07-23
fix: materialize missing `style`/`marks`/`markDefs` defaults without emitting patches

christianhg:
Opening a document where blocks lack `style` or `markDefs`, or spans lack `marks`, currently writes those defaults back into the document with the user's first edit. These fields are optional in Portable Text and every reader already defaults them, so this is a write nobody asked for.

The fix: the editor still fills the fields in its own memory, but no longer emits patches for them. The three normalization rules wrap their writes in `withoutPatching` (the same mechanism the placeholder-block rule already uses).

Why this is safe: nothing ever patches *against* these fields. They are only ever written whole (`set style "h1"`, `set marks ["strong"]`), so the editor and the document can disagree about a missing default without any patch misapplying. That's also the line that keeps `text` and `_key` emitting: text edits are dmp patches against the document's string, and `_key` is how patches address anything at all. And the renderers agree: `@portabletext/react`, `@portabletext/to-html`, and the toolkit all treat these fields as optional (`node.style || "normal"`, `markDefs ?? []`).

Two details worth knowing. Inserting a block still sends the complete block, defaults included, so new content reaches the document whole. And nineteen test pins expected the old default-`set` patches; all are updated by deleting exactly those patches from the expected output, nothing else changes. The deferred-normalization suite used these defaults as its test material, so a few of its scenarios now test less than their names suggest (flagged in a comment); they get rebuilt on `text`/`_key` in the pending-events queue rework.

### DIFF
--- a/.changeset/never-emit-default-materializations.md
+++ b/.changeset/never-emit-default-materializations.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: materialize missing `style`/`marks`/`markDefs` defaults without emitting patches
+
+The editor no longer writes default values for missing `style`, `marks`, or `markDefs` fields back to the document. These fields are optional in Portable Text with universal read-side defaults (renderers already default them), so the editor keeps its in-memory defaults and the document stays byte-identical to what its author wrote.
+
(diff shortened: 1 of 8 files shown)


christianhg:
Closing: wrong design. Never-emitting creates permanent engine-vs-document divergence on these fields, which breaks diff-based consumers (the SDK repair machinery, and the revised patch pipeline's own foundation, which requires the engine to byte-track the document). The right shape is to keep emitting these normalizations but flush them block-scoped, when a local edit touches the block, which is the pending-events queue rework. Decision recorded on the ticket.

PR #3017, opened 2026-07-23
fix(plugin-sdk-value): prevent transient duplicate remote pastes

ryanbonial:
## Description

Remote clipboard pastes can arrive as a temporary span insert followed by a text diff, while the cleanup unset reaches peers later. Receivers briefly rendered the pasted text twice.

Filter keyed inserted items that are absent from the SDK's current value before applying remote patches. The text diff still applies immediately.

## What to review

- `packages/plugin-sdk-value/src/plugin.sdk-value.tsx`: filters transient staged inserts from incoming remote patches.
- `packages/plugin-sdk-value/src/plugin.value-sync.browser.test.tsx`: covers the staged insert plus text-diff sequence.

## Testing

- `pnpm test:unit`
- `pnpm test:browser:chromium`
- `pnpm check:types`
- `pnpm check:lint`
- `pnpm build`
- Live two-browser clipboard-paste trace against the local build

## Notes for release

Prevents remote editors from briefly displaying pasted plain text twice.

### DIFF
--- a/.changeset/remote-paste-staging.md
+++ b/.changeset/remote-paste-staging.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/plugin-sdk-value': patch
+---
+
+fix: prevent transient duplicate text after a remote paste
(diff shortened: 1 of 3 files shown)


christianhg (approved):
Looks sensible! We can probably mitigate this in the editor engine by making it better at not producing scratch work. But that fix requires some more thought put into it and in the mean time this SDK mitigation makes sense to ship!

PR #3018, opened 2026-07-23
fix: cancel net-zero insert/unset pairs when flushing mutations

christianhg:
When someone pastes text into a paragraph, remote peers can briefly see the pasted text twice before it snaps back to normal. #3017 fixed this receiver-side for SDK consumers by filtering incoming inserts absent from the store's settled value; this PR fixes the sender so the transient content is never published in the first place, for every consumer.

The emission is the problem. A paste whose text merges into an existing same-mark span stages temporary spans and merges them away within one flush: the mutation carried all eight patches of that narration, two keyed inserts (the pasted span and an empty split remainder), the neighbor's `diffMatchPatch`, and two unsets retracting the inserts. Those pairs sum to nothing, and the store's final state is identical with or without them. But a receiver applies patches against its own engine state in the order delivered, and when delivery slices the transaction it renders the intermediate states. A probe with a `MutationObserver` on a receiving editor confirmed the split of responsibility: handed all the patches in one send, the DOM passes through exactly one state; handed them across sends, it paints the duplicate. The editor renders atomically per batch, so the only defense that works for every delivery is to not emit the scratch work.

`cancelNetZeroPatches` runs where a bulk flushes in the mutation batcher: a keyed item that is both inserted and unset within the flush is cancelled, together with every patch scoped inside it, and pairing is sequential so a re-insert of the same key survives as the net change. Two conservative guards keep it honest, pinned by unit tests over synthetic flushes: it bails per key when the key anchors another insert (cancelling an anchor target would dangle the reference) and when anything references the key outside the insert/unset window (a remove-then-recreate nets to a replacement, not to nothing). It assumes flush-created keys are fresh, which the key generator guarantees. A bulk whose patches all cancel is not emitted. The browser pin in `event.mutation.test.tsx` asserts the paste emission collapses to the single text diff, with the engine value agreeing.

Blast radius: document state is byte-identical by construction, undo is untouched (history is built from engine operations, not emitted patches), and individual `patch` events keep the full narration; only the `mutation` stream, which is what channels push to remotes, carries the collapsed form. All existing suites pass unmodified, editor and `plugin-sdk-value` alike, including the #3017 test: no existing pin ever asserted a mutation containing transient content. The one observable delta beyond fewer patches is that an edit whose patches fully cancel emits no mutation at all, which is the correct statement of "nothing changed".

### DIFF
--- a/.changeset/cancel-net-zero-mutation-patches.md
+++ b/.changeset/cancel-net-zero-mutation-patches.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: cancel net-zero insert/unset pairs when flushing mutations
+
+Edits that create and discard content within one flush, such as a paste whose text merges into an existing span, previously emitted `mutation` events carrying the scratch work: keyed inserts followed by the unsets retracting them. Receivers applying those patches in slices could transiently render the discarded content, for example a remote paste briefly showing its text twice.
+
+Emitted mutations now carry only the net change. Code observing `mutation` events sees fewer patches for such edits, and a mutation whose patches all cancel out is not emitted at all. Individual `patch` events are unchanged.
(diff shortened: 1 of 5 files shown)


christianhg:
Closing as part of the open-PR cleanup; the work is preserved on `cancel-net-zero-patches-at-mutation-assembly` and tracked in Linear for revival.

PR #3020, opened 2026-07-24
fix: position remote text patches relative to unflushed local edits

ryanbonial:
## Description

Two bodies of work, validated end-to-end against the concurrency safety harness (two live sessions on a real dataset).

### Position remote text patches relative to unflushed local edits

Two editors changing the same span concurrently could splice one editor's text into the middle of the other's. Incoming `diffMatchPatch` patches were fuzzy-matched against live text that already contained unresolved local edits, so repeated context could select the wrong position.

The editor now stores the text before unresolved local edits for each span. Incoming patches are matched exactly against that base and the live text. Base matches are merged with the local branch using deterministic three-way text merging. A live-only match proves that the sender incorporated the local change and retires the pending base.

Concurrent insertions at the same boundary use stable text ordering. Deletions remove only characters from the shared base, so concurrent insertions survive overlapping deletes. The stored base advances after each remote merge and is cleared on acknowledgement, structural replacement, value replacement, or expiry.

The remote patch batcher also preserves earlier successful patches when a later patch is ignored, so normalization and change notification still run.

### Converge concurrent editors on formatting, overlapping deletes, and emptied documents

Harness runs comparing this branch to the published packages surfaced three regressions: `delete-overlap` and `disjoint-blocks-format` went SAFE to SILENT-LOSS, and `link-overlap-range` persisted invalid Portable Text (an orphan mark with no `markDef`). Diagnosing with full value dumps and a Content Lake mutation listener showed three root causes, all fixed client-side:

- **Byte-level divergence on semantically identical values.** A span created locally serialized as `{_type, _key, ...}` while the same span adopted from a fetched document kept the server's alphabetized property order, so converged editors never became byte-identical. Nodes entering the engine are now rebuilt in a canonical property order (`_type`, `_key`, then the rest alphabetically) at the mutation chokepoints.
- **Empty documents could never converge.** Emptying the editor sets the field to `[]`, and every editor then re-minted placeholder keys on each adoption and repair pass. Adopting "empty" into an already-empty editor is now a no-op, clearing empties the first text block in place (keeping its keys, the same shape a select-all-delete leaves), and the plugin treats the local placeholder and an empty document as the same state instead of repairing one into the other forever.
- **markDefs housekeeping orphaned collaborators' marks.** Pruning unused definitions on value adoption deleted definitions that another client's in-flight edits still referenced, persisting spans whose marks had no definition. The pruning (unused definitions, duplicates, annotations on empty spans) now runs only as fallout of local edits, like span merging. The plugin additionally keeps every pushed transaction self-consistent: it never unsets a definition while the store, the editor, or the same flush still references it, re-inserts a definition the store lost when a flush references it, and removes stranded annotation references on confirmed repairs. Marks that never appeared in `markDefs` are untouched, so decorators from schemas this editor doesn't know still survive.

## Harness results (TRIALS=2, ROUNDS=3, real dataset)

| scenario | published | branch before | branch now |
|---|---|---|---|
| delete-overlap | SAFE | SILENT-LOSS 2/2 | SAFE 2/2 |
| disjoint-blocks-format | SAFE | SILENT-LOSS 1/2 | SAFE 2/2 |
| link-overlap-range | SILENT-LOSS 2/2 | CORRUPT 1/2 | SILENT-LOSS 1/2 |
| link-same-range | CORRUPT 1/2 | SILENT-LOSS 2/2 | SILENT-LOSS 2/2 |
| bold-overlap-range | SILENT-LOSS | SILENT-LOSS | SAFE 2/2 |
| block-split | SILENT-LOSS | SILENT-LOSS | SAFE 2/2 |
| paste-vs-type | SILENT-LOSS 1/2 | SILENT-LOSS 1/2 | SAFE 2/2 |

All typing scenarios stay SAFE, no scenario scores worse than published, and the link scenarios produced zero CORRUPT outcomes across 12 consecutive trials after the fix (previously roughly 1 in 3).

## What to review

- `packages/editor/src/internal-utils/span-text-merge.ts`: exact patch matching and deterministic three-way text merging.
- `packages/editor/src/internal-utils/applyPatch.ts`: base selection, acknowledgement, fuzzy fallback, and application to live text.
- `packages/editor/src/internal-utils/pending-local-text-edits.ts`: per-span base state and expiry.
- `packages/editor/src/editor/subscriber.local-text-edits.ts`: local, normalization, undo, redo, and structural lifecycle tracking.
- `packages/editor/src/engine/utils/canonicalize-properties.ts` plus its call sites in `apply-operation.ts` and `modify.ts`: canonical property order.
- `packages/editor/src/editor/sync-machine.ts`: key-preserving `clearEditor` and the empty-adoption no-op.
- `packages/editor/src/engine/core/normalize-node.ts` and `internal-utils/validateValue.ts`: markDefs housekeeping gated to local-edit fallout.
- `packages/plugin-sdk-value/src/plugin.sdk-value.tsx`: transaction-consistent markDefs decompose, stranded-mark healing, and empty-placeholder equivalence in the repair loop.

## Testing

- `pnpm test:unit` (editor): 889 passed
- `pnpm test:browser:chromium` (editor): 1,761 passed, 1 skipped
- `plugin-sdk-value` unit: 52 passed; chromium browser (incl. collision matrix and multi-round link overlap): 58 passed
- `pnpm check:types`, `pnpm check:lint`: passed in both packages
- Harness: full 19-scenario matrix run on the fixed build (table above), plus TRIALS=4 confidence runs on the link scenarios

Updated tests that pinned replaced behaviors: placeholder re-minting on clear (keys are now preserved), unused-markDef pruning on adoption (now kept until the next local edit), one literal property-order assertion, the multi-round link test (editors now converge on the server value verbatim), and the plugin's empty-store tests (no synthetic caret).

## Notes for release

Fixes concurrent edits to the same span being merged inside each other or losing characters, concurrent formatting and overlapping deletes leaving sessions permanently diverged, and concurrent link formatting persisting marks without a matching `markDef`. Sessions now converge on the same deterministic result, and documents stay valid Portable Text.


christianhg:
Closing as part of the open-PR cleanup of parked work; the branch `fix/concurrent-paste-typing-race` preserves both bodies of work and the harness validation notes, and the effort is tracked in Linear for revival. Closing is bookkeeping, not a verdict on the approach.

PR #3027, opened 2026-07-29
fix: prevent the native event when a matched Behavior's actions return nothing

### LINKED ISSUE #2271: `actions: [() => []]` doesn't call preventDefault despite guard match
### Description

When a behavior's guard matches but all action functions return empty arrays (`[]`), the native browser event is **not** prevented — even though the behavior has claimed ownership of the event.

This creates a contradiction: `defaultBehaviorOverwritten` is set to `true` (the behavior "owns" the event and no other behaviors run), but `nativeEventPrevented` remains `false` (the browser's default action still fires).

### Reproduction

```typescript
// This prevents the browser default (correct):
defineBehavior({
  on: 'keyboard.keydown',
  guard: ({ event }) => event.originEvent.key === 'Tab',
  actions: [],
})

// This does NOT prevent the browser default (unexpected):
defineBehavior({
  on: 'keyboard.keydown',
  guard: ({ event }) => event.originEvent.key === 'Tab',
  actions: [() => []],
})
```

In the second case, the guard matches, `defaultBehaviorOverwritten` is set to `true` in `behavior.perform-event.ts`, and the `break` after the action loop prevents any other behavior from handling the event. But because the action function returned `[]`, the `continue` in the action loop skips it without setting `nativeEventPrevented = true`.

The result: the behavior claimed the event (no other behavior runs), but the browser's default action (e.g., Tab changing focus) still fires.

### Expected behavior

If a behavior's guard matches, the native event should be prevented regardless of what the action functions return. The behavior has claimed ownership — an empty action return should mean "I handled it by doing nothing" (noop), not "actually never mind, let the browser have it."

### Suggested fix

After the action loop completes for a matched behavior, if `defaultBehaviorOverwritten` is `true` but `nativeEventPrevented` is still `false`, set `nativeEventPrevented = true`. This ensures that a matched guard with empty action returns still suppresses the browser default, while preserving the existing `forward` semantics (which explicitly sets `nativeEventPrevented = false` when it wants the native event to proceed).

### Impact

This is a subtle API footgun. `actions: []` and `actions: [() => []]` look nearly identical but have opposite effects on `preventDefault`. Developers writing conditional action logic (where an action function might return `[]` in some cases) will hit unexpected browser behavior.

Real-world example: a Shift+Tab behavior at indent level 0 that wants to do nothing. If written as `actions: [({snapshot}) => { if (level === 0) return []; ... }]`, the browser's default Tab behavior fires at level 0 even though the guard matched.

voidhrithik:
Fixes #2271.

```diff
     // This Behavior now "owns" the event and we can consider the default
     // action prevented
     defaultBehaviorOverwritten = true

-    if (eventBehavior.actions.length === 0) {
-      nativeEventPrevented = true
-    }
+    nativeEventPrevented = true
```

The comment above it already said this. Claim the event as soon as the guard matches, and leave the assignment inside the action loop as the only thing that can lower it again, which is how `forward` already gets its way.

I tried the post-loop fix from the issue first. It fixes the reported case and breaks `forward`, because once the loop has exited, false-because-forward and false-because-nothing-ran are the same value and the override flips both.

Tests are in `packages/editor/tests/behavior.native-event-prevented.test.tsx`. Three cases, each pressing Tab for real and reading `defaultPrevented` off the event once it reaches `document`. On unfixed `main` the `[() => []]` case fails and the other two pass. On the post-loop version the `forward` case fails instead. That third test passes on `main`, so on its own it proves nothing, it is only there to catch the override.

Browser suite 1752 passing, unit 872, lint and types clean.

One behaviour change worth flagging: an action set that throws leaves `actions` empty, so it now prevents where `main` did not. I think that is right, since a Behavior that just crashed should not hand the keystroke back to the browser, but it is a judgement call and I am happy to special-case it.

### DIFF
--- a/.changeset/olive-pugs-repeat.md
+++ b/.changeset/olive-pugs-repeat.md
@@ -1,0 +1,5 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: prevent the native event when a matched Behavior's actions return nothing
--- a/packages/editor/src/behaviors/behavior.perform-event.ts
+++ b/packages/editor/src/behaviors/behavior.perform-event.ts
@@ -219,9 +219,7 @@ export function performEvent({
     // action prevented
     defaultBehaviorOverwritten = true
 
-    if (eventBehavior.actions.length === 0) {
-      nativeEventPrevented = true
-    }
+    nativeEventPrevented = true
 
     let actionSetIndex = -1
 
--- a/packages/editor/tests/behavior.native-event-prevented.test.tsx
+++ b/packages/editor/tests/behavior.native-event-prevented.test.tsx
@@ -1,0 +1,113 @@
+import {describe, expect, test} from 'vitest'
+import {userEvent} from 'vitest/browser'
+import {forward} from '../src/behaviors/behavior.types.action'
+import {defineBehavior} from '../src/behaviors/behavior.types.behavior'
+import {BehaviorPlugin} from '../src/plugins/plugin.behavior'
+import {createTestEditor} from '../src/test/vitest'
+
+const initialValue = [
+  {
+    _type: 'block',
+    _key: 'k0',
+    children: [{_type: 'span', _key: 'k1', text: 'foo'}],
+  },
+]
+
+/**
+ * Records `defaultPrevented` for the first Tab keydown that reaches `document`.
+ * Listening on `document` in the bubble phase means the editor's own handler on
+ * the editable element has already run by the time this fires.
+ */
+function recordTabPrevented() {
+  const seen: Array<boolean> = []
+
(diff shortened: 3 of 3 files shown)


voidhrithik:
@christianhg could you approve the CI run when you get a chance? It's behind the first-time-contributor gate, so nothing has actually run yet.

One thing worth flagging before you read the diff: I didn't use the fix you suggested in the issue. It breaks `forward`, and one of the three tests fails on it. Reasoning is in the description.

christianhg (approved):
Good stuff! Thank you, @voidhrithik ❤️

PR #3034, opened 2026-08-03
fix: auto-resolve duplicate `markDefs` at value ingress

christianhg:
Extracted from #3033 as its own mechanism: a new `validateValue` auto-resolution, not part of the cleanup gating or the emission fix. Based on #3033's branch so the diff shows only this commit; GitHub retargets to `main` when #3033 merges.

Duplicate `markDef` keys are structural, not cosmetic: a patch addressing `markDefs[_key=="x"]` can hit either copy. A value arriving with duplicates is now repaired on arrival, first copy wins, via a whole-array `set` (keyed unsets would themselves be ambiguous against the duplicates). Without this, a document that already carries duplicates keeps them in-engine until a local edit touches that block; with it, the ambiguity never enters the session.

Two honest scope notes. Auto-resolutions never surface in Studio's invalid-value UI (they apply as a `console.warn` plus ordinary patches), so the new i18n keys are inert metadata and no Studio version dependency exists. And the repair's outbound write follows the existing auto-resolve semantics: on `update value` it emits and heals the stored document; on initial load it repairs the engine's copy while the write parks, the same first-load behavior as the long-standing orphaned-`markDefs` resolution.

Pinned: `pteWarningsSelfSolving` gains the dedupe repair scenario, and the gated-normalization suite's adoption pin re-homes from "keeps duplicates as-is" to "repairs at ingress", with the on-touch scenario building its duplicate state through a remote patch, the one path that (deliberately) bypasses ingress.

### DIFF
--- a/.changeset/dedupe-markdefs-ingress.md
+++ b/.changeset/dedupe-markdefs-ingress.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: repair duplicate `markDefs` when a value arrives
+
+A document arriving with the same mark definition twice (same `_key`) is repaired on arrival: the first copy wins, and the repair is written back with the user's first edit. Duplicate keys made patch addressing ambiguous, so unlike the deferred cleanup rules this one is treated as a validity repair.
--- a/packages/editor/src/internal-utils/validateValue.ts
+++ b/packages/editor/src/internal-utils/validateValue.ts
@@ -230,6 +230,40 @@ export function validateValue(
           ),
         ]
 
+        // Duplicate `markDef` keys make keyed addressing ambiguous (a patch
+        // targeting `markDefs[_key=="x"]` can hit either copy). The repair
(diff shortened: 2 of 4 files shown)


christianhg:
Closing unmerged. Review surfaced three problems with the repair's whole-array `set` emission: it can clobber a collaborator's concurrent definition insert on the raw patch channel (an unprompted write racing user edits); through the SDK plugin it heals nothing, the mergeable decomposition keys the store by `_key`, which collapses the duplicates and reduces the repair to zero operations; and with a dirty store it risks a pull-repair/ingress-dedupe cycle that nothing tests.

The world without this PR is sound: the editor mirrors dirty documents faithfully (no divergence), #3033's upsert prevents new duplicates, and the first local edit to an affected block heals it through the established, user-anchored emission path. Ingress-repair semantics for definition tables belong in the revised-patch-pipeline design, where write shape and concurrency can be decided together.

PR #3036, opened 2026-08-03
feat(plugin-sdk-value): show remote carets and report presence

ryanbonial:
## What

Presence in a Portable Text field. Other people's carets show up where they are editing, and the local user's caret is reported so they show up for everyone else, including in the Studio.

`SDKPortableTextEditable` replaces `PortableTextEditable` and wires up all of it, so the document and field are named once:

```tsx
<EditorProvider initialConfig={{schemaDefinition}}>
  <SDKPortableTextEditable documentId={id} documentType="post" path="content" />
</EditorProvider>
```

That's value sync, presence reporting, and visible carets. A separate `SDKValuePlugin` is no longer needed alongside it.

## Why it lives here, and why one component

Presence started as a separate `plugin-sdk-presence` package. It moved here because splitting it made the developer pass the same handle and path twice, once to the editable and once to the value plugin, for two features that almost always ship together.

The single-component shape is what makes presence automatic, and it isn't about packaging. `rangeDecorations` is a prop on `PortableTextEditable` with no plugin API behind it, so a sibling plugin can never draw carets on its own. The Studio has the same constraint and solves it the same way: `PortableTextInput` owns the render and concatenates its presence decorations with the caller's. This does that, so `rangeDecorations` you pass are kept and merged rather than replaced.

Alternatives rejected:

- **Cursor hook in `@sanity/sdk-react`.** `RangeDecoration` is an editor type, so the SDK would need `@portabletext/editor` as a peer while this package needs the SDK. A mutual peer dependency across two separately released repos.
- **A plugin API for range decorations in `@portabletext/editor`.** The better long-term answer, and it would help diff decorations and search highlighting too, but the decorations actor is created inside `Editable` from the prop, so it's a real API change rather than something a plugin can do. Worth its own issue.
- **Keeping presence in its own package.** Cleaner in principle, worse in practice: a presence-only consumer would still install `xstate`, `@sanity/diff-patch`, and `@sanity/json-match` for value-sync machinery it never uses.

## Breaking-ish

The `@sanity/sdk-react` peer range moves from `^2.1.2` to `^2.19.0`, which is where the presence hooks were added. Existing installs are untouched; npm consumers upgrading this package will need to bump the SDK too, and pnpm/yarn will warn.

## What to review

- `src/presence-cursors.ts` is the whole model, pure and framework-free: collapse a selection to a caret, compare selections by value, resolve where each caret draws, record moves, build decorations.
- `src/sdk-editable.tsx` is the wrapper. The prop split is the risky part, since handle fields must not reach the DOM. `UnhandledHandleKeys` is a compile-time assertion that fails if `DocumentHandle` gains a field. It already earned its place: `@sanity/sdk-react`'s `DocumentHandle` extends the core one with `resourceName`, which I'd missed. Note also that `resource` is both a handle field and an RDFa HTML attribute, so the handle wins via `Omit`.
- `src/presence-caret.tsx` is the built-in caret, ported from Studio's `UserPresenceCursor` without `@sanity/ui`, styled-components, or motion. Two structural details are load-bearing: the caret is a zero-width sibling before `children`, not a wrapper, and it needs `contentEditable={false}` or the local user can put their own cursor inside someone else's caret.
- `src/selection-compatibility.test.ts` pins that the SDK's wire selection and `EditorSelection` stay mutually assignable, since the plugin passes values through in both directions rather than converting.

## Testing

96 unit tests and 183 browser tests across Chromium, Firefox, and WebKit. The pre-existing value-sync and collision-matrix suites are unchanged and still pass.

Mutation tested throughout. Ten mutations of the caret model and wrapper logic are all caught. Two initially survived and both were my own bad tests: a "same colour in two tabs" test that called the colour function twice and asserted nothing about the component, and no coverage of the wrapper defaulting to the built-in caret. The first is now a browser test rendering two sessions of one user plus a second user and comparing computed colours; the second is covered by extracting `resolveCursorRenderer` and testing all three cases.

Also verified by hand against a real socket, running the SDK kitchensink's Portable Text route against this branch in two browser tabs.

One thing worth knowing if you test it: two editors in the same tab cannot see each other. Presence stores are keyed only by project and dataset in a module-level registry, so they share one store and one session id, and a session ignores its own reports. Use two tabs, or a tab plus a Studio.

### DIFF
--- a/packages/plugin-sdk-value/src/index.ts
+++ b/packages/plugin-sdk-value/src/index.ts
@@ -1,1 +1,13 @@
+export {
+  SDKPresencePlugin,
+  useSDKPresenceCursors,
+  type RenderCursorFunction,
+  type SDKPresencePluginProps,
+  type SDKRemoteCursor,
+  type UseSDKPresenceCursorsOptions,
+} from './plugin.sdk-presence'
 export {SDKValuePlugin, ValueSyncPlugin} from './plugin.sdk-value'
+export {
+  SDKPortableTextEditable,
+  type SDKPortableTextEditableProps,
+} from './sdk-editable'
--- a/packages/plugin-sdk-value/src/presence-cursors.ts
+++ b/packages/plugin-sdk-value/src/presence-cursors.ts
@@ -1,0 +1,159 @@
+import type {EditorSelection, RangeDecoration} from '@portabletext/editor'
+import {isEqualSelections} from '@portabletext/editor/utils'
+import type {PropsWithChildren, ReactElement} from 'react'
+
+/**
+ * A remote participant's position in the editor.
+ *
+ * @public
+ */
+export interface RemoteCursor {
+  /**
+   * Identifies one editing session rather than one person. The same user in two
+   * tabs reports two sessions and therefore draws two carets.
+   */
+  sessionId: string
+  /**
+   * The selection the participant last reported, or `null` when they have none.
+   */
+  selection: EditorSelection
+}
+
+/**
+ * Where the local editor has pushed a remote caret, and the reported selection
+ * it was pushed from.
+ *
+ * Typing above or inside a remote caret moves it, and the editor reports the
+ * new position through `onMoved`. Keeping the selection it moved from is what
+ * lets a later report from that participant take over again.
+ *
+ * @public
+ */
+export interface CursorOverride {
+  reported: EditorSelection
+  current: EditorSelection
+}
+
+/**
+ * Options for {@link toRangeDecorations}.
+ *
+ * @public
+ */
+export interface RangeDecorationOptions<TCursor extends RemoteCursor> {
+  cursors: readonly TCursor[]
+  overrides: ReadonlyMap<string, CursorOverride>
+  /**
+   * Builds the component that draws one caret. The plugin has no opinion about
+   * how a caret looks, so this is the caller's to provide.
+   */
+  renderCursor: (cursor: TCursor) => (props: PropsWithChildren) => ReactElement
+  onCursorMoved?: (cursor: TCursor, selection: EditorSelection) => void
+}
+
+/**
+ * Reduces a selection to a caret at its focus point.
+ *
+ * Presence answers where someone is, so decorating their whole selection would
+ * highlight text the local user never selected. The Studio collapses the same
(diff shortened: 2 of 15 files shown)


packages/plugin-sdk-value/src/presence-cursors.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Compares two selections by value.
+ *
+ * Reported selections arrive as fresh objects every time, so comparing by
+ * reference would read an unchanged caret as a move.
+ *
+ * @public
+ */
+export function isEqualSelection(
```
christianhg: I realise this one is slightly different than the one exported from `@portabletext/editor/utils`, but maybe we can reuse (and/or adjust) the one exported from PTE?

packages/plugin-sdk-value/src/index.ts
```diff
@@ -1 +1,29 @@
...
+  toRangeDecorations,
+  type CursorOverride,
+  type RangeDecorationOptions,
+  type RemoteCursor,
+} from './presence-cursors'
+export {getCaretColor, renderDefaultCursor} from './presence-caret'
+export {
+  SDKPortableTextEditable,
+  type SDKPortableTextEditableProps,
+} from './sdk-editable'
```
christianhg, line 13: I wonder if the amount of exported functions here is more-than-needed. Is there an opportunity here to trim experts that are not strictly needed by consumers? (Always easier to add stuff like this than remove later)

christianhg (approved):
Don't have the capacity to do a full review, but this feature is definitely valuable! And I agree with the single-component compromise. But maybe trim the exported surface before merging?

packages/plugin-sdk-value/src/index.ts
ryanbonial: ooh yep. trimmed from 21 new exports down to 8:

```
SDKPortableTextEditable, SDKPortableTextEditableProps
SDKPresencePlugin, SDKPresencePluginProps
useSDKPresenceCursors, UseSDKPresenceCursorsOptions
RenderCursorFunction, SDKRemoteCursor
```

Most of the rest were only exported because the tests import them. Dropped the pure cursor helpers and their types, plus some others.

PR #3038, opened 2026-08-04
chore: remove the root `CLAUDE.md`

christianhg:
The file froze in December. Its command list duplicates what any agent discovers from `package.json` scripts, and its style bullets have drifted behind what review actually enforces. The rename history (`CLAUDE.md` → `AGENTS.md` twice, then back once with a copy-pasted commit subject) is the churn signature of a file nobody owns, and a stale instruction file is worse than none: agents obey it.


christianhg:
Superseded by the rename-and-rewrite: a current AGENTS.md beats deletion.

PR #3040, opened 2026-08-04
refactor: drive normalization from an ordered rule table

christianhg:
Answering "what rules does the editor uphold, and when do they run" today requires reading a 652-line `normalizeNode` walker with ~17 rules as ordered if-blocks, eight inline remote-change guards, and a prose enumeration of the gating policy in `types/editor-engine.ts` that turned out to have drifted from the implementation — proof that a non-executed description of these rules rots.

Each rule is now a named function in an ordered, module-private `NORMALIZE_RULES` table (`{id, runsDuringRemoteChanges, normalize}`) that the walker executes, so the manifest cannot drift. The gates hoist into the table column; `text-block.adjacent-spans` is `'partial'`, keeping its arm-level split (merge gated, inline-object bracketing not). The rotten comment now points at the table. Nothing is exported: rule ids would freeze as public surface mid-flux, and nothing registers rules dynamically, so there is no registration API.

The ids double as a shared defect vocabulary: the overlapping checks in `validateValue` and `parse-blocks` are tagged with keep-in-sync comments naming the same ids and both entrance policies (comment-only). The verified policy disagreements between validation and normalization are deliberately untouched — surfaced as vocabulary, audited separately.

Zero semantic delta, pinned four ways: the `[id, gate]` pairs as literal data, rule order (fails on a row swap, executed), runtime gate enforcement including the `'partial'` arm split, and span pass-through emitting nothing in both modes. Full unit (879) and chromium (1,764) suites green with identical counts.

### DIFF
--- a/packages/editor/src/engine/core/normalize-node.test.ts
+++ b/packages/editor/src/engine/core/normalize-node.test.ts
@@ -1,0 +1,227 @@
+import {
+  compileSchema,
+  defineSchema,
+  type PortableTextBlock,
+  type PortableTextTextBlock,
+} from '@portabletext/schema'
+import {createTestKeyGenerator} from '@portabletext/test'
+import {describe, expect, test} from 'vitest'
+import {createEditor} from '../create-editor'
+import {normalize} from '../editor/normalize'
+import type {Editor} from '../interfaces/editor'
+import type {Node} from '../interfaces/node'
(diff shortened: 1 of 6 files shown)


christianhg:
Closing unmerged: superseded by a redrawn plan (typed defect-id vocabulary from day one, staged across all three rule systems). The branch stays for reference.

PR #3042, opened 2026-08-05
chore: bump `@sanity/pkg-utils` to v12

stipsan:
Bumps the tooling catalog from `@sanity/pkg-utils` `^10.8.2` to `^12.1.0` (tsdown build stack) and migrates package configs to the v12 API.

## Build times

Clean `pnpm --filter './packages/*' build` on Node 24.19.0:

| | `@sanity/pkg-utils` | Wall time |
| --- | --- | --- |
| **Before** | 10.8.2 | **54.1s** |
| **After** | 12.x | **41.2s** |

About **1.3× faster** (−13s). The "after" figure is the mean of two clean runs (41.16s, 41.35s).

## Changes

- Catalog: `@sanity/pkg-utils` → `^12.1.0`.
- Migrate `package.config.ts` to the v12 API: `babel`/`reactCompilerOptions` → `reactCompiler`, `extract` → `tsdoc`, drop `dts: 'rolldown'`.
- Drop `--clean` from the `pkg-utils build` scripts, which is the default in v12.
- Remove the `"jsx": "preserve"` override from every package `tsconfig.settings.json`. v12 removed pkg-utils' own `jsx` option and reads the setting from `tsconfig` instead, and `@vitejs/plugin-react` v6 likewise delegates to Vite's oxc pipeline, so `preserve` left raw JSX in the built `lib`/`dist` output and in the vitest transforms. `@sanity/tsconfig/recommended` already sets `react-jsx`.
- Raise Vite to `^8.2.0` and `@vitejs/plugin-react` to `^6.0.5` — the peer chain `reactCompiler` pulls in through `@sanity/tsdown-config` — and bump `@tailwindcss/vite` accordingly. Vite and vitest configs move to `reactCompilerPreset` plus `@rolldown/plugin-babel`, since plugin-react v6 dropped its `babel` option.
- **`@portabletext/editor` stops declaring `./test` and `./test/vitest` exports.** Neither has ever resolved from the published package — both conditions pointed at `src` files the tarball does not contain, and v12's publint check rightly rejects that. They are monorepo-only test helpers, so rather than start publishing them (which would have forced `vitest`, `vitest-browser-react`, `racejar`, and `@portabletext/test` into optional `peerDependencies`), they leave `exports` entirely. In-repo consumers map the two specifiers through tsconfig `paths` to the editor's source entry files, and every consumer's vitest projects reuse that same mapping via Vite 8's `resolve.tsconfigPaths`. Consumers' type-check tsconfigs widen `rootDir` to the packages directory because their program now contains editor sources; builds are unaffected since `tsconfig.dist.json` extends the settings file directly and pins its own `rootDir`. With the test entries no longer built, `@textspec/notation` also drops out of the editor's `inlinedDependencies`.
  - Note for a follow-up: `apps/docs/src/content/docs/editor/guides/testing-behaviors.mdx` still tells readers to import from `@portabletext/editor/test/vitest`, which has never worked from npm and now is explicitly monorepo-only.
- `@portabletext/plugin-table` stops publishing `src`: the stylesheet moves through the v12 CSS pipeline into `dist`, with `css: {target: false}` so Lightning CSS keeps `light-dark()` intact instead of lowering it into undefined custom properties.
- Changesets cover the published manifest changes: every package gains a `module` field, `@portabletext/test` stops publishing a `source` condition that points at unpublished `src`, the editor drops the never-resolvable `./test` entry points, and plugin-table's stylesheet moves to `dist`.

## Why `bundle-stats` is red

**`bundle-stats` is expected to fail on this PR and to pass again once it merges to `main`.** No action is needed on the branch.

`rexxars/bundle-stats@v1` compares head against base by checking out the *baseline* commit and building it — but using the dependencies this PR installed. So pkg-utils 12 ends up evaluating `main`'s pre-bump `package.config.ts` files and throws:

```
error Error: package.config.ts: `dts: 'rolldown'` was removed in @sanity/pkg-utils v12
```

for `@portabletext/patches`, `racejar`, `@portabletext/keyboard-shortcuts`, and `@portabletext/schema`. The giveaway that this is `main`'s tree and not ours is the build command in those log lines: `pkg-utils build --strict --check --clean`, still carrying the `--clean` flag this PR removes. The head build itself succeeds; only the baseline half of the comparison fails, so no treemap artifacts are produced and the action exits 1.

Any breaking major of the build tool hits this, because the pre-bump tree cannot be built with the post-bump toolchain. Once this lands on `main`, both sides of the comparison use the v12 config API and the check works again. The only ways to force it green now would be workflow edits — skipping the baseline comparison, or setting `NODE_ENV=production` so pkg-utils skips its legacy-config validation — which would be changing CI purely to silence a check, so I left it failing.

Separately, `Browser tests (chromium)` fails intermittently with `[vitest] Browser connection was closed while running tests`, aborting the editor suite partway with no assertion failures. That predates this PR (it happened on `normalize-rule-table` on Aug 4 under Vite 7 and pkg-utils 10), lands on a different test file every time, and passes on rerun. It is worth its own issue rather than a change here.

<div><a href="https://cursor.com/agents/bc-2e067386-b58d-4acf-8a8f-e6afde86b489?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-2e067386-b58d-4acf-8a8f-e6afde86b489&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


.changeset/brave-moons-invite.md
```diff
@@ -0,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: publish `./test` and `./test/vitest` entry points that resolve
+
+Both entry points pointed at `./src/test/…` in `exports` and
+`publishConfig.exports` alike, against `files: ["lib"]`, so they resolved to
+files the tarball does not contain. They are built now and resolve to `lib`.
```
stipsan: @christianhg hmmm it looks like these entrypoints were never intended to be published to npm?

stipsan: I can see they are currently, but they won't resolve, so it would be better if the entrypoints didn't make it to npm right? 

Image

PR #3043, opened 2026-08-05
refactor: describe normalizations as corrections in an ordered store

christianhg:
Answering "what corrections does the editor make to a document, and when" today means tracing 652 lines of `normalizeNode` control flow: ~17 repairs as ordered if-blocks applying operations imperatively mid-walk, with eight inline remote-change gates deciding which of them run on remote fallout.

A correction is now a set of operations, nothing more. `corrections.ts` describes 15 of the repairs as data — `{name, type, correct: (context) => operations | undefined}` in one ordered array — and `normalizeNode` becomes the executor. Scheduling derives from the classification: `structural` corrections (the engine cannot hold the document otherwise) run on every walk, `cosmetic` ones (canonicalization of valid content) only as local-edit fallout, which reproduces the old gates exactly. Definition metadata carries what a pure repair cannot: patch suppression for the empty-editor placeholder, and the verified-unique-keys memo, generalized as `memoKey`. Nothing is exported; the store is closed by design.

Two repairs stay imperative at their original positions — the same-mark span merge and the per-child merge/drop/bracket loop — because they depend on `applyMergeNode`'s ref pre-transformation and mid-loop refetching. They convert when `merge` becomes a first-class `EngineOperation` (next increment).

The first commit is the guardrail: an op-stream equivalence harness (52 op-stream assertions across 14 defect fixtures, in both local and remote mode, generated against `main` and verified green there) passes unchanged after the restructure. Full unit (893) and chromium (1,764) suites green with identical counts. Three non-API deltas are disclosed in the commit body: per-rule debug strings became correction names, one provably unreachable check is dropped, and two explicit early-returns are replaced by the type-exclusivity invariant that made them redundant.

### DIFF
--- a/packages/editor/src/engine/core/corrections.test.ts
+++ b/packages/editor/src/engine/core/corrections.test.ts
@@ -1,0 +1,214 @@
+import {
+  compileSchema,
+  defineSchema,
+  type PortableTextBlock,
+} from '@portabletext/schema'
+import {createTestKeyGenerator} from '@portabletext/test'
+import {describe, expect, test} from 'vitest'
+import {createEditor} from '../create-editor'
+import {normalize} from '../editor/normalize'
+import type {Editor} from '../interfaces/editor'
+import {CORRECTIONS} from './corrections'
+
+const schema = compileSchema(defineSchema({}))
+
+type Fixture = Record<string, unknown>
+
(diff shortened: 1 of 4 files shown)


christianhg:
Closing as part of the open-PR cleanup; the corrections-store work is preserved on `corrections-store` and tracked in Linear for revival.

PR #3044, opened 2026-08-05
fix: re-resolve the caret when remote patches move, absorb, or split off its span

christianhg:
Put your caret after the `b` in "foo**bar**" and let a collaborator unbold "bar": their editor merges the now-identical spans, the merge crosses the wire as a text patch plus an `unset`, and your caret jumps to the end of the word. The same collapse hits a collaborator's Backspace joining two blocks (with same marks the sender also span-merges in the same batch, so the caret lands in the previous block at a meaningless offset) and a collaborator's Enter before your caret, which clamps a caret sitting in the moved tail to the truncated block's end instead of following it into the new block. Sanity patches carry state deltas, not position mappings, so by the time the batch arrives the edit's intent is gone and the generic per-op selection transforms can only collapse the caret to a boundary.

The fix recovers the position instead of inferring the intent. Before applying a remote batch, `setupRemotePatches` captures the caret's span `_key` and text, its offsets (in the span and in the block's text), and the adjacent sibling blocks. Afterwards, recovery runs only inside the neighborhood a structural edit can actually reach: a key-and-text match in the captured block or its previous sibling is followed; a dead span whose block text is unchanged, a dead block whose previous sibling now holds exactly the concatenated text, or a truncated block whose freshly inserted next sibling holds exactly the removed tail each re-resolve through `blockOffsetToSpanSelectionPoint`. Anything inexact keeps the generic result.

The neighborhood scoping is load-bearing: splitting a block reuses the span `_key` in the new block, so real documents routinely contain doc-wide duplicate span keys and a doc-wide key match proves nothing; a dedicated scenario pins that a remote block delete never follows the caret's key to an unrelated same-keyed span. One deliberate policy change is included: the existing split scenario pinned the clamp at the truncated block's end, and a caret inside the moved tail now follows its content into the new block. Each recovery is pinned red-first, including a two-editor scenario in which a real `insert.break` on one editor relays its actual mutation patches to the other; the remaining scenarios pass unchanged.

### DIFF
--- a/.changeset/remote-merge-caret-recovery.md
+++ b/.changeset/remote-merge-caret-recovery.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: re-resolve the caret when remote patches move, absorb, or split off its span
+
+When a collaborator's edit merges the span the caret sits in (for example unbolding text so adjacent spans join), merges its block into the previous one with Backspace, or splits the block before the caret with Enter, the caret now keeps its text position instead of jumping to a span boundary. In the split case the caret follows the content it sat in into the new block; previously it stayed clamped at the end of the truncated block. Recovery is scoped to the caret's block and its adjacent siblings and only applies when the text mapping is exact; in every other case, including duplicate span keys elsewhere in the document (which block splits produce routinely), the previous behavior stands.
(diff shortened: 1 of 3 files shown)


christianhg:
Closing as part of the open-PR cleanup; the caret-recovery work is preserved on `remote-caret-recovery` and tracked in Linear for revival.

PR #3048, opened 2026-08-06
chore(deps): bump vite, vitest deps, @vitejs/plugin-react, and playwright

stipsan:
Bumps the tooling catalog's `vite`, `@vitejs/plugin-react`, and `@playwright/test`, migrating off `@vitejs/plugin-react`'s removed inline `babel` option along the way. `vitest`, `@vitest/browser`, `@vitest/browser-playwright`, and `vitest-browser-react` were already at latest, so no change there.

## Bumps

- `vite`: `^7.3.5` -> `^8.2.0`
- `@vitejs/plugin-react`: `^5.2.0` -> `^6.0.5`
- `@tailwindcss/vite`: `^4.1.18` -> `^4.3.3` (its peer range doesn't accept `vite@8` until this major)
- `@playwright/test`: `^1.60.0` -> `^1.62.1` (root `package.json` and `packages/racejar/package.json`, both hardcoded rather than cataloged)

`@sanity/pkg-utils@10.8.2` (the package build tool) has no dependency on `vite` at all, so this bump doesn't touch the package build pipeline (`package.config.ts`, `dts: 'rolldown'`) — only vite/vitest dev & test configs and the playground/example apps.

## The `@vitejs/plugin-react` v6 breaking change

v6 dropped its bundled Babel dependency and the inline `babel` option ([release notes](https://github.com/vitejs/vite-plugin-react/releases/tag/plugin-react%406.0.0)). Every config wiring up `babel-plugin-react-compiler` through that option:

```ts
react({babel: {plugins: [['babel-plugin-react-compiler', {target: '19'}]]}})
```

now uses the exported `reactCompilerPreset` helper with a separate `@rolldown/plugin-babel` plugin, following the same fix as [portabletext/editor#3042](https://github.com/portabletext/editor/pull/3042) (scoped here to just this dependency bump, not that PR's `@sanity/pkg-utils` v12 migration):

```ts
import babel from '@rolldown/plugin-babel'
import react, {reactCompilerPreset} from '@vitejs/plugin-react'
// ...
plugins: [react(), babel({presets: [reactCompilerPreset({target: '19'})]})]
```

Any `include`/`exclude` filtering of *which files the compiler runs on* moves to the `babel()` call too, since that's the plugin that now actually runs `babel-plugin-react-compiler` (`react()`'s own `exclude` only gates its Fast Refresh transform). Added `@rolldown/plugin-babel` to the tooling catalog and as a `devDependency` (`dependencies` for the private `playground` app, matching #3042) everywhere this pattern was used: `apps/playground`, `examples/basic`, `packages/editor`, `packages/toolbar`, and the `plugin-{typography,typeahead-picker,table,sdk-value,paste-link,markdown-shortcuts,list-index,dnd,input-rule,emoji-picker,character-pair-decorator}` packages.

Two follow-up cleanups to that filtering, both dead code carried over verbatim from the pre-migration configs (unrelated to the version bump itself, just made visible by moving the filter around):
- `apps/playground/vite.config.ts`'s `engine-dom`/`engine-react` exclude patterns never matched anything — the engine's `dom`/`react` code lives nested under `src/engine/dom/`/`src/engine/react/`, not sibling `src/engine-dom/`/`src/engine-react/` directories, and the remaining `/src/engine/` pattern already covers everything nested under it.
- `packages/editor/vitest.config.ts`'s exclude referenced the pre-rename `slate`/`slate-dom`/`slate-react` names, which don't exist either (renamed to `engine` the same way). Updated to `/src/engine/` and moved it onto the `babel()` plugin so the compiler actually skips the vendored engine code again during tests.

## The `"jsx": "preserve"` breaking change

v6 delegates JSX transformation to Vite's oxc pipeline, which respects the project's tsconfig `jsx` setting. Every package above overrode it to `"preserve"` in `tsconfig.settings.json` (needed for `@sanity/pkg-utils`' own Babel-based build step, unaffected by this bump since it doesn't depend on `vite`), which left raw JSX in the vitest transform and broke browser tests. Removed the override everywhere `@vitejs/plugin-react` is used in a vite/vitest config, falling back to `@sanity/tsconfig`'s `react-jsx` default. Packages with the same override that don't run `@vitejs/plugin-react` in tests (`racejar`, `plugin-one-line`, `markdown`, `html`, `keyboard-shortcuts`) are untouched.

## No changeset

Prior `vite`/`vitest`/`@vitejs/plugin-react`/`@playwright/test` catalog and peerDependency bumps (including racejar's `@playwright/test` peer range) never shipped a changeset — these are dev/tooling-only, so the same applies here.

## Testing

- `pnpm install` resolves cleanly; `pnpm exec playwright install chromium` re-pulled the browser for 1.62.1.
- `pnpm check:format`, `check:lint`, `check:types`, `check:react-compiler`, `check:knip` all pass.
- `pnpm --filter @portabletext/editor build`: succeeds; spot-checked `lib/index.js` for correctly transformed JSX (no raw JSX left over from the `tsconfig.settings.json` change).
- `pnpm test:unit` and `pnpm test:browser:chromium` (full monorepo, then re-verified for `@portabletext/editor` after the `babel()` exclude fix): all green.
- Manually smoke-tested `pnpm dev:playground` on Vite 8 — typing, selection, and Bold formatting all work, no console errors.

Note: CI on this PR is currently blocked by an ongoing [GitHub Actions outage](https://www.githubstatus.com) unrelated to this change (workflow runs aren't starting at all repo-wide); it'll pick up the latest commit once GitHub recovers.

[Playground smoke test on Vite 8 with bold text applied](https://cursor.com/agents/bc-d3d9e31d-3204-4352-a244-3321f8d5e625/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fplayground_smoke_test_vite8_bold.webp)

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#team-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-d3d9e31d-3204-4352-a244-3321f8d5e625?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d3d9e31d-3204-4352-a244-3321f8d5e625&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/apps/playground/package.json
+++ b/apps/playground/package.json
@@ -31,6 +31,7 @@
     "@portabletext/plugin-typography": "workspace:*",
     "@portabletext/react": "catalog:tooling",
     "@portabletext/toolbar": "workspace:*",
+    "@rolldown/plugin-babel": "catalog:tooling",
     "@xstate/react": "catalog:",
     "emojilib": "^4.0.2",
     "fuse.js": "^7.1.0",
--- a/apps/playground/vite.config.ts
+++ b/apps/playground/vite.config.ts
@@ -1,23 +1,16 @@
 import path from 'node:path'
+import babel from '@rolldown/plugin-babel'
 import tailwindcss from '@tailwindcss/vite'
-import react from '@vitejs/plugin-react'
+import react, {reactCompilerPreset} from '@vitejs/plugin-react'
 import {defineConfig} from 'vite'
 
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [
-    react({
-      babel: (id) => {
-        const isVendoredEngine =
-          id.includes('/src/engine/') ||
-          id.includes('/src/engine-dom/') ||
-          id.includes('/src/engine-react/')
-        return {
-          plugins: isVendoredEngine
-            ? []
-            : [['babel-plugin-react-compiler', {target: '19'}]],
-        }
-      },
+    react(),
+    babel({
+      exclude: [/[/\\]node_modules[/\\]/, /[/\\]src[/\\]engine[/\\]/],
+      presets: [reactCompilerPreset({target: '19'})],
     }),
     tailwindcss(),
   ],
--- a/examples/basic/package.json
+++ b/examples/basic/package.json
@@ -18,6 +18,7 @@
   },
   "devDependencies": {
     "@eslint/js": "^9.34.0",
+    "@rolldown/plugin-babel": "catalog:tooling",
     "@types/react": "catalog:tooling",
     "@types/react-dom": "catalog:tooling",
     "@vitejs/plugin-react": "catalog:tooling",
--- a/examples/basic/vite.config.ts
+++ b/examples/basic/vite.config.ts
@@ -1,14 +1,11 @@
 import path from 'node:path'
-import react from '@vitejs/plugin-react'
+import babel from '@rolldown/plugin-babel'
+import react, {reactCompilerPreset} from '@vitejs/plugin-react'
 import {defineConfig} from 'vite'
 
 // https://vite.dev/config/
 export default defineConfig({
-  plugins: [
-    react({
-      babel: {plugins: [['babel-plugin-react-compiler', {target: '19'}]]},
-    }),
-  ],
+  plugins: [react(), babel({presets: [reactCompilerPreset({target: '19'})]})],
   resolve: {
     alias: {
       '@portabletext/editor': path.resolve(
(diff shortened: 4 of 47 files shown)


christianhg (approved):
Thank you ❤️

PR #3066, opened 2026-08-11
chore(markdown): update @portabletext/toolkit to v6

rexxars:
Updates `@portabletext/markdown` to `@portabletext/toolkit@6.0.0`. It is the only package in this repo that depends on the toolkit.

## The list fix does not affect this package

The headline change in toolkit v6 is to `nestLists()`, which fills in list levels that were never authored so nesting depth always matches `level`. **This package never calls `nestLists()`.** Markdown expresses list depth through indentation, so `DefaultListItemRenderer` reads each block's `level` directly and emits `'   '.repeat(level - 1)`.

All 292 tests pass unchanged, `tsc` and `biome lint` are clean.

## Node version

The toolkit has required Node.js 22.12 since v6, so `engines.node` for `@portabletext/markdown` goes from `>=20.19 <22 || >=22.12` to `>=22.12`. Advertising support for Node.js 20.19 while depending on a package that requires 22.12 was incorrect, and Node.js 20 reached end of life in April 2026.

**Only `@portabletext/markdown` is changed.** The other 22 packages here still declare `>=20.19 <22 || >=22.12`. They all currently share one range, which suggests they are kept in lockstep, so bringing the rest along is a repo-wide call rather than something to fold into this PR. Happy to do it as a follow-up if that is the direction.

## Note on `pnpm-workspace.yaml`

`pnpm add` wanted to reformat the whole file, switching every quoted key from double to single quotes and dropping the `# Renovate security update: sharp@0.35.0` comment. I reverted that and added the single catalog line by hand, so the diff is one line.

## Unrelated pre-existing bug found while testing

Not fixed here, and not caused by this change, but worth recording.

Round-tripping a list that starts deeper than level 1, or that skips levels, is lossy. Because the indent is computed from the absolute `level` rather than the depth relative to the list actually emitted, `portableTextToMarkdown` produces indentation that CommonMark reads as something else.

Input: a `number` item at level 3, then level 1, then a `bullet` at level 1, a level 4 item, and back to level 1. Output:

```
      1. Starts at level 3
1. Back to level 1
- Bullet level 1
         - Jumps to level 4
- Bullet level 1 again
```

Feeding that straight back through `markdownToPortableText`:

- `Starts at level 3` has a 6 space indent with no preceding list, so it parses as an **indented code block** and stops being a list item at all
- `Jumps to level 4` has a 9 space indent, so it is absorbed into the previous item as literal text, `"Bullet level 1\n- Jumps to level 4"`, rather than becoming a nested item

This is the same class of problem as portabletext/toolkit#100, just in a different renderer: a level jump greater than one is not translated into a single step of nesting. The fix would be to indent by the depth actually emitted rather than by `level - 1`. Let me know if you want that as a separate PR.

### DIFF
--- a/.changeset/nine-pears-repeat.md
+++ b/.changeset/nine-pears-repeat.md
@@ -1,0 +1,9 @@
+---
+"@portabletext/markdown": major
+---
+
+Node.js 22.12 or later is now required
+
+`@portabletext/toolkit` has required Node.js 22.12 since v6, so continuing to advertise support for Node.js 20.19 here was incorrect. Node.js 20 reached end of life in April 2026.
+
+If you build on Node.js 20, upgrade to 22.12 or later before taking this version.
--- a/.changeset/tricky-pots-shake.md
+++ b/.changeset/tricky-pots-shake.md
@@ -1,0 +1,7 @@
+---
+"@portabletext/markdown": patch
+---
+
+Update `@portabletext/toolkit` to v6
+
+Nothing changes in the markdown output. The headline fix in v6 is to `nestLists()`, which this package does not use, since markdown expresses list depth through indentation and reads each block's `level` directly.
--- a/packages/markdown/package.json
+++ b/packages/markdown/package.json
@@ -55,7 +55,7 @@
   "dependencies": {
     "@mdit/plugin-alert": "^0.23.2",
     "@portabletext/schema": "workspace:^",
-    "@portabletext/toolkit": "^5.0.2",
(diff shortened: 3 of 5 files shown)


rexxars:
Superseded. The toolkit v6 bump is not needed here: `@portabletext/markdown` never calls `nestLists()`, and bumping it alone would take this package's dependency range and Node floor out of sync with the other 22 packages in the monorepo.

Reopening the useful half separately, as a fix for the round-trip bug found while testing this.

PR #3068, opened 2026-08-11
fix(markdown): serialize lists that start deep or skip levels

rexxars:
Fixes lists that start deeper than level 1, or skip levels, serializing to Markdown that does not read back as a list. Same class of bug as portabletext/toolkit#100, in a different renderer.

## The bug

`DefaultListItemRenderer` indented each item by its absolute `level`, three spaces per level:

```ts
const indent = '   '.repeat(level - 1)
```

Markdown cannot express depth that way. Indentation is relative to the list item above, and a first item indented four spaces or more is an indented code block, not a list.

Feeding a `number` item at level 3, then level 1, then a `bullet` at level 1, one at level 4, and one back at level 1 through `portableTextToMarkdown` produced:

```
      1. Starts at level 3
1. Back to level 1
- Bullet level 1
         - Jumps to level 4
- Bullet level 1 again
```

Reading that straight back with `markdownToPortableText` lost two of the five items:

- `Starts at level 3` has a six space indent with no preceding list, so it parsed as a **code block** and stopped being a list item
- `Jumps to level 4` has a nine space indent, so it was absorbed into the previous item as literal text, `"Bullet level 1\n- Jumps to level 4"`

## The fix

`buildListIndexMap` now also computes the depth each item should render at. Every jump to a deeper level counts as **one** step of nesting, however many levels it spans, and returning to a shallower level pops back to the matching ancestor. The default renderer indents by that depth instead of by `level`.

List numbering is keyed off the same depth, so items rendered into one list are numbered as one sequence. Keying it off `level` while rendering by depth would have let two items sit in the same visual list with independent counters.

Same input now:

```
1. Starts at level 3
2. Back to level 1
- Bullet level 1
   - Jumps to level 4
- Bullet level 1 again
```

All five items survive the round-trip, at levels 1, 1, 1, 2, 1.

## Why collapse rather than pad

The toolkit fixes this in HTML by generating the missing levels as empty `<li>` wrappers, because HTML has no other way to nest a list. Markdown does not have that constraint, and padding would mean emitting empty `- ` bullets into what is meant to be human-editable source. Collapsing produces exactly what someone would write by hand. The trade-off is that absolute `level` values are normalised on the round-trip, which seems right given a gap is usually an authoring artifact.

## Scope

- Lists that start at level 1 and change one level at a time are **unaffected**. All 292 existing tests pass untouched, plus 5 new ones.
- Custom list item renderers get a new optional `listDepth` prop with the depth to indent by. `value.level` is unchanged, so existing custom renderers keep working exactly as before.
- **`packages/plugin-list-index` is deliberately left alone.** It holds a duplicate of the engine's index computation, with a drift alarm in its test suite saying to update both together. That does not apply here: the editor indents by `level`, so keying its numbering off `level` is correct there. Only Markdown needs the collapse. Flagging it so nobody re-syncs the two by reflex.

## Verification

- `@portabletext/markdown`: 297 tests pass, up from 292
- `@portabletext/plugin-list-index`: 32 pass
- `@portabletext/editor` unit: 1375 pass, no type errors
- `tsc --noEmit` and `biome lint` clean

## Not included

This deliberately does **not** bump `@portabletext/toolkit` to v6, which was the original reason I was in here. `@portabletext/markdown` never calls `nestLists()`, so v6 changes nothing for it, and bumping this package alone would take its dependency range and Node floor out of sync with the other 22 packages in the monorepo. That is better done as a repo-wide sweep. See the closed #3066.

### DIFF
--- a/.changeset/great-jars-repeat.md
+++ b/.changeset/great-jars-repeat.md
@@ -1,0 +1,27 @@
+---
+"@portabletext/markdown": patch
+---
+
+Fix lists that start deeper than level 1, or skip levels, serializing to broken Markdown
+
+`portableTextToMarkdown` indented each list item by its absolute `level`, at three spaces per level. Markdown cannot express depth that way: indentation is relative to the list item above, and a first item indented four spaces or more is a code block, not a list.
+
+A list item at level 3 with nothing above it produced six spaces of indent, so reading it back gave a code block instead of a list item, losing the item. A jump from level 1 to level 4 produced nine spaces, which was absorbed into the previous item as literal text rather than becoming a nested item.
+
+Each jump to a deeper level is now a single step of nesting, however many levels it spans, so the output is valid Markdown that reads back as the same list structure. List numbering follows the list an item is actually rendered into, rather than its `level`.
+
+Given a `number` item at level 3 followed by one at level 1, then a `bullet` at level 1, one at level 4, and one back at level 1:
+
+```md
+1. Starts at level 3
+2. Back to level 1
+- Bullet level 1
+   - Jumps to level 4
+- Bullet level 1 again
+```
+
+Previously the first line was indented into a code block and `Jumps to level 4` was swallowed into the line above it.
+
+Lists that start at level 1 and change one level at a time are unaffected.
+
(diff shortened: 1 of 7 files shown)


christianhg (approved):
Thank you ❤️

PR #3072, opened 2026-08-14
feat!: require node 22.12 and drop the deprecated v7 surface

christianhg:
## What

The v8 major: every commit is a breaking change that could only ship in a major, one commit per item, each with its own changeset. All 23 published packages bump together.

**Node 22.12 floor.** Node 20 reached end of life in April 2026, and the first-party dependency family already moved (`@portabletext/react` v8, `to-html` v6, `toolkit` v6, all Node-floor majors). `engines` narrows from `>=20.19 <22 || >=22.12` to `>=22.12` everywhere, and the editor and markdown packages adopt `to-html` v6 and `toolkit` v6, which carry the same floor and no API changes.

**The DOM speaks `data-pt-*` only.** The Slate-era `data-slate-*` aliases are gone; the changeset carries the full old→new attribute table. One behavior change rides along: a node registered via `registerNode` now receives identical clean attributes whether its parent text block renders through the legacy or the new pipeline; the mode-inheritance tests pin the surrounding legacy DOM (`data-child-*`) as the mode signal instead.

**`renderListItem` and core list indices are gone.** The render prop was the only reason core computed list numbering; the prop, the `data-list-index` attribute on legacy blocks (`data-list-item`/`data-level` stay), the lazy `listIndexMap`, and its selector channel all delete together. `@portabletext/plugin-list-index` is the replacement and its `buildListIndexMap` is now the sole implementation instead of a keep-in-sync duplicate.

**`renderBlock` and `renderChild` are gone as well.** With Studio's comment input migrated to registrations (its last consumer of either prop), the final two block-level render props delete, with their function and props-shape types and the `LegacyRenderHooks` threading. The legacy default rendering for unregistered nodes is byte-identical, verified attribute by attribute in review; only the callback slots disappear. The prop test suites port to registrations with their contracts intact, the playground's `RenderBlock` becomes a catch-all `defineBlockObject`, and the rendering guides rewrite to the registration-only reality. Studio's comment-input migration is already merged, so its next editor bump compiles clean.

**The built-in `text/markdown` converter is gone.** It made every consumer bundle `markdown-it` (~123KB minified) and serialized markdown on every copy, for a clipboard entry almost nothing reads; real markdown arrives as `text/plain`. Pastes carrying `text/markdown` fall back to `text/html`/`text/plain` (a converter-less mime type dead-ends the pipeline, so the priority entry goes too). `plugin-table` keeps its pinned GFM table copy by producing `text/markdown` itself, and the migration guide carries a restoration snippet; there is no public converter-registration seam, so nothing else can depend on the removed converter.

**`renderStyle` is gone too.** Text-block styles render through `defineTextBlock`, which already owns the wrapper wherever a text block is registered. With both props removed, the playground, the basic example, the docs demo editor, and the getting-started tutorials all migrate to registered text blocks (list attributes re-emitted via `useListIndex`); because a registered text block also suppresses `renderChild` for its inline children, their stock tickers move to `defineInlineObject`. The `BlockListItemRenderProps` and `BlockStyleRenderProps` types are removed as well, in their own commit further down, once Studio stopped typing against them. Both props were deprecated in v7 (#3075); the deprecation asides that PR put on the docs pages disappear across this PR's commits as each prop's removal lands, and none survive at the tip: nothing deprecated remains to warn about.

**The deprecated `value` mirror and `AddedAnnotationPaths` fields are gone.** Compiled style, list, and decorator schema types no longer carry `value` (read `name`; nothing typed against the mirror, Studio's `.value` reads target `@sanity/types` definitions where `value` is canonical). `addAnnotation` returns only `markDefPaths`; Studio's one consumer of `markDefPath` migrates in a small Studio PR; nothing here gates on it, Studio bumps to v8 on its own schedule.

**`resolveContainerAt` is unexported.** The positional container resolver shipped `@alpha` with the container redesign and found no consumers anywhere. It stays internal; `getContainerChildren` remains the public per-node primitive.

**One default DOM.** Unregistered nodes stop emitting the legacy dressing (`pt-*` classes, `data-block-*`, `data-child-*`, `data-style`, `data-list-item`, `data-level`, the root's `pt-editable` class) and render the same clean `data-pt-*` shape registered defaults produce. Behavior stays: the `contentEditable`/`draggable` wrappers, `dir`, spacers, `[type: key]` previews, and `data-read-only`. The built-in drop indicator is removed along with its now-orphaned render store; drop chrome is `@portabletext/plugin-dnd`'s job, which already tracks positions from the public `drag.*` events, while drop semantics (`behavior.core.dnd`) stay in core. Studio has zero usage of any removed hook, verified by code search. The shortcut plugins also drop their deprecated `schema`/`level` callback params in their own majors.

**Dropping a block mid-paragraph now splits it.** Internal whole-block drags were the one path snapping to before/after while paste and external drops already split at the caret; the drag placement now resolves the same way, degenerating to snapping at block edges, on voids, on inline-object leaves, and when caret resolution falls back to an expanded selection (where splitting would delete content). `plugin-dnd` hides its edge indicator during mid-block hovers, the native drop caret is the affordance there.

**Deprecated `loading`, `done loading`, and `error` events removed** from `EditorEmittedEvent`, keeping their JSDoc promise. Exhaustive switches over `event.type` lose three cases.

**Legacy `main`/`module` fields dropped**; `exports` is the single entry surface, matching `toolkit` v6. pkg-utils builds and publint stay green.

A non-breaking rider removes the last Slate references and repairs six upstream issue links that a past rename had pointed at a nonexistent `ianstormtaylor/engine` repo.

## Deliberately kept

The `PortableTextEditor` static class, `usePortableTextEditor`, and the dual render pipeline's default rendering stay: Studio still consumes the statics and hooks throughout its PT input, so their removal needs a coordinated Studio migration first. The invalid-value gate also stays; its userland extraction is sequenced separately.

Verified with the full editor browser suite (chromium), all unit suites, and every root check CI runs. One known-red check: `bundle-stats` rebuilds `main` as its baseline against this branch's install state, and `main`'s editor still declares the removed `@portabletext/markdown` workspace dependency, so the baseline pack cannot resolve it. The check self-heals on merge; the alternative (a planted devDependency keeping the symlink alive) is worse than the red.


christianhg:
Fully replaced by the stack; closing. The ledger, commit by commit:

**Merged as standalone PRs**: node floor #3106, `data-slate-*` #3108, relay events #3109, `main`/`module` #3110, `renderListItem` #3111, `renderStyle` #3114, `renderBlock`/`renderChild` #3116, mark props #3123 (scope addition, not on this branch), props-shape types #3126, `resolveContainerAt` #3131, slate refs #3132, unified DOM + orphaned store #3138, split-on-drop #3140, markdown converter #3144 (redesigned en route: fail-forward paste pipeline, `text/markdown` kept as protocol). Vanilla example open as #3149.

**Deliberately not shipped in v8**: the compiled-schema `value` mirror (merged then reverted, #3130), `markDefPath`/`spanPath` (#3125 closed, leaves with `addAnnotation` in v9), and the shortcut plugins' `schema`/`level` params (EDEX-1996 dropped). All three wait on v9 behind runtime deprecation warnings (#3134).

The branch stays for v9 reference: it carries the three dropped removals as ready commits.

PR #3077, opened 2026-08-14
refactor: route default type renderer refusal through a `guarded` classifier

christianhg:
The markdown serializer dispatches on the `_type` name alone, so a default type renderer receives arbitrary values whenever a consumer schema declares a same-named type of a different shape. The table default (the only default type renderer today) handles this with a classify-then-render pair: `isTableShaped` gates between the typed `renderTable` and the fenced-JSON `DefaultUnknownTypeRenderer` fallback. Before more default renderers ship and each improvises its own version, this PR extracts the wiring into one internal `guarded(isShaped, render)` combinator and makes the table renderer its first caller.

The combinator carries the pattern's one rule in its contract: the predicate must check exactly what the render function dereferences, no more and no less, so the narrowing is earned rather than asserted. `guarded<V, T>` returns `PortableTextTypeRenderer<V>` with the render callback typed against `V & T`; `V` infers from the annotation at the assignment site, which keeps the compiler link between `DefaultTableRenderer`'s public annotation and its implementation that the hand-wired version had. No cast, no `any` seam.

`guarded` stays internal. Making refusal consumer-visible (renderers returning `undefined` with the dispatcher owning the fallback) is a real API design with its own trade-offs and is deliberately not decided here.

Pure refactor: zero observable change on every dispatch path (shaped table → GFM, unshaped value → fenced JSON, empty `rows` → empty string), pinned by the existing suite. No changeset.

Stacked on #3073; retarget to `main` after it merges.

### DIFF
--- a/packages/markdown/src/from-portable-text/renderers/type.ts
+++ b/packages/markdown/src/from-portable-text/renderers/type.ts
@@ -5,7 +5,10 @@ import {
   escapeImageAndLinkTitle,
   escapeTableCell,
 } from '../../escape'
-import type {PortableTextTypeRenderer} from '../types'
+import type {
+  PortableTextTypeRenderer,
+  PortableTextTypeRendererOptions,
+} from '../types'
 
 /**
  * @public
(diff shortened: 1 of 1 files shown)


christianhg:
Closing unmerged: persona-counting against the package's actual objective showed the combinator is speculative abstraction. The safety it generalizes (default renderers must not crash on `_type` collisions) already ships inside #3073's `DefaultTableRenderer`, and the remaining default renderers need one-to-two-line inline checks each, not shared machinery. The sibling renderers (react-portabletext, to-html) build no guard framework either: components trust, and only library defaults verify. The rule survives as prose on the follow-up ticket.

PR #3085, opened 2026-08-17
feat: split text blocks when a block is dropped inside them

christianhg:
Dropping a block object in the middle of a paragraph snapped it to before or after the paragraph, while the browser's drag caret pointed inside it. Paste and external drops already split the paragraph at the caret, so internal block drags were the one path that behaved differently. They now resolve the same way: the drop splits the paragraph, and drops at block edges, on voids, on inline objects, and with unresolved caret positions keep snapping like before.

The restructuring surfaced a data loss bug that is fixed here too: dropping blocks that the destination rejects (a table cell that accepts none of them, for example) deleted the drag source with nothing inserted. That drop is now a no-op.

The built-in drop indicator and `plugin-dnd`'s edge indicator both hide while hovering inside a text block, since the browser's own drop caret is the affordance there, and a boundary line would promise a before/after placement the drop won't make. Edge lines still show for boundary, void, and unresolved positions, and a line painted by an earlier boundary hover clears when the pointer moves mid-block.

Nine new drop scenarios pin the behavior, including the split itself (fails on the old code), the rejected-drop no-op, table-cell splits, and one-step undo.

### DIFF
--- a/.changeset/fix-empty-fit-drop.md
+++ b/.changeset/fix-empty-fit-drop.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: keep the drag source when a drop fits no blocks
+
+Dropping blocks that the destination rejects entirely (a table cell whose schema accepts none of them, for example) deleted the drag source with nothing inserted. That drop is now a no-op.
--- a/.changeset/plugin-dnd-mid-block-indicator.md
+++ b/.changeset/plugin-dnd-mid-block-indicator.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/plugin-dnd': patch
+---
+
+fix: hide the edge indicator while hovering inside a text block
+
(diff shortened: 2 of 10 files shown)


christianhg:
Closing per review discussion: the split feature folds into the v8 major, where the dual-chrome complexity it had to manage on main does not exist, and the data-loss guard's premise turned out to be unproven (its pinning test passes on unguarded code). The guard ships separately if a scripted reproduction with a real drag flow confirms the loss; the investigation is tracked on the data-integrity ticket.

christianhg:
Correction to the close comment: the data-loss premise was true after all. The pinning test had been silently appended as an empty string (a marker-collision bug in the extraction script), so it never ran; with the test actually in place, unguarded code fails with the dragged block gone from the value. The fix ships in isolation as #3086 with that red-on-old proof. The split feature still folds into v8 as decided.

PR #3125, opened 2026-08-20
feat!: remove `markDefPath` and `spanPath` from `AddedAnnotationPaths`

christianhg:
`annotation.add` reported three paths; `markDefPaths` (one per selected block, in selection order) is the canonical one, and Studio's annotation flows read `markDefPaths[0]`. The single-block `markDefPath` and the never-consumed `spanPath` are removed.

Part of the v8 stack.

---

> [!NOTE]
> **Medium Risk**
> Major semver break for consumers still reading markDefPath or spanPath; runtime addAnnotation behavior is slightly stricter about when paths are returned.
> 
> **Overview**
> **Breaking:** `PortableTextEditor.addAnnotation` / `EditableAPI.addAnnotation` now return only **`markDefPaths`** on `AddedAnnotationPaths`; deprecated **`markDefPath`** and **`spanPath`** are removed.
> 
> The editable implementation no longer picks a single focus-span `markDef` for `markDefPath`. It still returns paths when the focus block/span exists and the focus span gained at least one of the newly created mark defs, with **`markDefPaths`** listing every new mark def in document order (for multi-block selections, use **`markDefPaths[0]`** where the old single-block `markDefPath` applied). A major changeset is included for `@portabletext/editor`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b7a8448c4edbadd6674d175a522d049986fdd185. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/remove-added-annotation-paths-fields.md
+++ b/.changeset/remove-added-annotation-paths-fields.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': major
+---
+
+feat!: remove `markDefPath` and `spanPath` from `AddedAnnotationPaths`
+
+`PortableTextEditor.addAnnotation` now returns only `markDefPaths` (document order). `markDefPath` was the focus-span's markDef and `spanPath` was documented as not meaningful. Consumers pick from `markDefPaths`: the first entry matches the old `markDefPath` whenever the annotation covers one block.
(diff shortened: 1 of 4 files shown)


christianhg:
Dropped by decision: `addAnnotation` is itself deprecated and leaves with the `PortableTextEditor` statics in the next major, so removing deprecated fields from its return type is churn on a dying surface. The fields ride out with the whole API.

PR #3134, opened 2026-08-21
fix: warn at runtime when the deprecated `value` mirror or legacy callback params are read

christianhg:
v8 keeps two deprecated compat surfaces by decision: the `value` mirror on compiled schema entries and the top-level `schema`/`level` params on the shortcut plugins' callbacks. Their JSDoc deprecation reaches only consumers who typecheck; the configs that would break on a future removal are untyped and would fail silently. This PR makes reading either surface log a one-time `console.warn` naming the replacement (`name`, `context.schema`, `props.level`) and the next-major removal horizon, so the v9 removal lands on consumers who were warned at runtime.

The `value` mirror becomes an enumerable accessor with the old contract intact: spread, serialization, and deep equality see it unchanged, and assignment still works (the setter redefines it as plain data). Wrapping copies property descriptors instead of spreading, so the library's own internal re-wrapping neither warns to itself nor burns the once-flag before a consumer's first read; a production-path test pins zero warnings through `compileSchema` and `getSubSchema` until a consumer actually reads `value`. The callback params become getters on the argument object; destructuring counts as a read and keeps working.

One commit per package, each with a patch changeset. Part of the v8 stack; the removals these warnings precede are next-major scope.

### DIFF
--- a/.changeset/warn-deprecated-character-pair-decorator-callback-arg.md
+++ b/.changeset/warn-deprecated-character-pair-decorator-callback-arg.md
@@ -1,0 +1,10 @@
+---
+'@portabletext/plugin-character-pair-decorator': patch
+---
+
+
+fix: warn once when reading the deprecated `schema` callback argument
+
+The `decorator` callback still receives the deprecated top-level `schema` argument, but reading it now logs a one-time console warning naming `context.schema` as the replacement. A callback that only reads `context.schema` sees no warning. The deprecated argument is removed in the next major.
+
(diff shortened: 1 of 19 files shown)


christianhg:
Closed by decision: v8 keeps the deprecated surfaces as plain fields, and their v9 removal rides on JSDoc deprecation and the major bump alone. The getter-and-warn machinery bought runtime reach into untyped configs at the cost of accessor semantics on hot compiled objects and three packages of warn-once plumbing; not worth it for the population involved.

PR #3149, opened 2026-08-24
chore: add a zero-config vanilla example

christianhg:
A full content model (styles, lists, decorators, a link annotation, an image block object, a mention inline object) with no render props, no registrations, no plugins, and no CSS: what the editor's own defaults render and what they don't (object previews as `[type: key]` tokens, lists with no visual treatment, no UI for anything the schema allows). `pnpm --filter example-vanilla dev` serves it.

The last piece of the v8 stack.

### DIFF
--- a/examples/vanilla/index.html
+++ b/examples/vanilla/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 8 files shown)


christianhg:
Dropped by decision: the playground and `examples/basic` already cover the demo surface, and a third runnable target is maintenance without an audience. The zero-config teaching it carried lives on the Rendering concepts page (what the engine defaults do and don't render).

PR #3157, opened 2026-08-24
fix: identify shadow roots by node type instead of a `host` property

### LINKED ISSUE #3155: containsShadowAware infinite-loops (locks the tab) when a form contains an element with id="host" — DOM clobbering
### Summary

`containsShadowAware` and `closestShadowAware` detect a `ShadowRoot` with `'host' in current.parentNode`. Because `HTMLFormElement` exposes its own controls as **named properties**, any form containing an element with `id="host"` (or `name="host"`) makes `'host' in form` evaluate to `true`. The upward walk then steps `form → form.host` (the input) → back up to `form` → `form.host` → … forever.

The result is a hard lock of the browser tab: the main thread never returns, CPU pins at 100–190%, and Chromium eventually kills the renderer ("Aw, Snap!"). It reproduces in Chrome and Edge, in normal and incognito windows, with no extensions.

This is a textbook [DOM clobbering](https://en.wikipedia.org/wiki/DOM_clobbering) issue: `'host' in node` is not a safe ShadowRoot test.

### Affected code

`packages/editor/src/engine/dom/utils/dom.ts` — two occurrences on `main`:

```ts
// closestShadowAware
} else if (current.parentNode && 'host' in current.parentNode) {
  current = (current.parentNode as ShadowRoot).host as DOMElement
```

```ts
// containsShadowAware
if ('host' in current.parentNode) {
  current = (current.parentNode as ShadowRoot).host
}
```

### Minimal reproduction (no editor involved)

This isolates the walk itself. Run in any Chromium browser:

```js
const walk = (parent, child, limit = 10_000) => {
  if (parent.contains(child)) return 'early-true'
  let current = child
  let steps = 0
  while (current && steps++ < limit) {
    if (current === parent) return 'true'
    if (current.parentNode) {
      current = 'host' in current.parentNode ? current.parentNode.host : current.parentNode
    } else {
      return `terminated cleanly after ${steps} steps`
    }
  }
  return '*** INFINITE LOOP ***'
}

const detached = document.createElement('span')

document.body.innerHTML = '<form><input id="host"><input id="other"></form>'
console.log(walk(detached, document.getElementById('other')))
// -> "*** INFINITE LOOP ***"

document.body.innerHTML = '<form><input id="smtpHost"><input id="other"></form>'
console.log(walk(detached, document.getElementById('other')))
// -> "terminated cleanly after 5 steps"
```

Note the trigger is the *presence* of `id="host"` anywhere in the form — not the node you start from — because every walk that reaches the form bounces off `form.host`.

### Real-world reproduction (Sanity Studio)

Sanity derives an input's DOM `id` from the field name, so a document type with a field named `host` renders `<input id="host">` inside the document form.

1. Define a document type with a field `host` (type `string`) and at least one Portable Text field.
2. Open a document of that type in the Studio.
3. Click into one Portable Text editor, then click into another one.
4. The tab locks; CPU pins at 100%+.

Renaming the field to anything other than `host` fixes it immediately. Interrupting the hung renderer with the debugger gives:

```
#0 containsShadowAware
#1 setDomSelection
#2 (anonymous)                    <- useIsomorphicLayoutEffect
#3 react_stack_bottom_frame
#4 runWithFiberInDEV
#5 commitHookEffectListMount
#6 commitHookLayoutEffects
#7 commitLayoutEffectOnFiber
#8 recursivelyTraverseLayoutEffects
   ... (repeating)
```

`host` is a natural field name (SMTP/database/server configuration), so this is easy to hit accidentally, and the symptom gives no hint of the cause.

### Suggested fix

Test for a real `ShadowRoot` instead of a property named `host`. A `ShadowRoot` is a `DocumentFragment` (`nodeType === 11`); an `HTMLFormElement` is `nodeType === 1`, so a nodeType check is enough and stays safe when the `ShadowRoot` global is absent (SSR, some test environments):

```ts
const isShadowRoot = (node: Node): node is ShadowRoot =>
  node.nodeType === 11 /* DOCUMENT_FRAGMENT_NODE */ && 'host' in node
```

Using `instanceof ShadowRoot` alone would also fix the clobbering, but breaks across realms (iframes) and where the global is undefined.

Opened as #3157, with the guard applied to both helpers plus regression tests.

### Environment

- `@portabletext/editor` 7.11.0 (also present on `main` at time of writing)
- Reproduced via Sanity Studio 6.10.1
- Chrome and Edge (Chromium), macOS 15 arm64; normal and incognito windows, no extensions
- Does **not** reproduce in headless Chromium, because the hang needs a real focus/selection change to drive `setDomSelection`

tillschweneker:
Fixes #3155.

### The bug

`closestShadowAware` and `containsShadowAware` identified a shadow root with `'host' in current.parentNode`. `HTMLFormElement` exposes its own controls as named properties, so `'host' in form` is `true` for **any** form containing an element with `id="host"` or `name="host"`.

Such a form was then treated as a shadow root and the walk stepped `form` → `form.host` (the control) → back up to `form` → `form.host` → … . It never terminates: the main thread hangs, CPU pins at 100–190%, and the browser kills the tab.

Sanity Studio derives an input's DOM `id` from the field name, so a document type with a field named `host` (SMTP/server config — a very natural name) renders `<input id="host">` inside the document form. Clicking between two Portable Text fields on such a document locks the tab, because `setDomSelection` calls `containsShadowAware` from a layout effect:

```
#0 containsShadowAware
#1 setDomSelection
#2 (anonymous)                    <- useIsomorphicLayoutEffect
#3 react_stack_bottom_frame
#4 runWithFiberInDEV
#5 commitHookEffectListMount
#6 commitHookLayoutEffects
#7 commitLayoutEffectOnFiber
#8 recursivelyTraverseLayoutEffects
   ... (repeating)
```

### The fix

Identify shadow roots by node type. Shadow roots are document fragments; a form is an element, so the two no longer collide:

```ts
const isShadowRoot = (value: any): value is ShadowRoot => {
  return (
    isDOMNode(value) &&
    value.nodeType === Node.DOCUMENT_FRAGMENT_NODE &&
    'host' in value
  )
}
```

The node type is checked rather than `instanceof ShadowRoot` so the guard keeps working across realms (iframes) and in environments where the global is undefined. Both call sites now use it.

### Tests

`packages/editor/tests/dom-shadow-aware.test.ts` covers both helpers: ordinary containment, shadow-boundary crossing (still works), and the regression — a form owning a control named `host`.

The tests live in the browser project rather than the unit project on purpose: **jsdom does not implement form named-item access**, so `'host' in form` is `false` there and the bug cannot be reproduced under the unit runner. I confirmed that before moving them.

One caveat worth flagging: against the pre-fix code these tests do not fail, they **hang** — the loop is synchronous, so it blocks the very thread Vitest would use to time the test out. That is the nature of the bug rather than a property of the tests. If you would prefer a bounded-iteration guard so CI fails fast instead of stalling, I am happy to add one.

For a runner-independent demonstration, this reproduces the loop with the pre-fix logic in any Chromium browser:

```js
const walk = (parent, child, limit = 10_000) => {
  if (parent.contains(child)) return 'early-true'
  let current = child
  let steps = 0
  while (current && steps++ < limit) {
    if (current === parent) return 'true'
    if (current.parentNode) {
      current = 'host' in current.parentNode ? current.parentNode.host : current.parentNode
    } else {
      return `terminated cleanly after ${steps} steps`
    }
  }
  return '*** INFINITE LOOP ***'
}

const detached = document.createElement('span')

document.body.innerHTML = '<form><input id="host"><input id="other"></form>'
console.log(walk(detached, document.getElementById('other')))
// -> "*** INFINITE LOOP ***"

document.body.innerHTML = '<form><input id="smtpHost"><input id="other"></form>'
console.log(walk(detached, document.getElementById('other')))
// -> "terminated cleanly after 5 steps"
```

### Checks run

Per `AGENTS.md`: `pnpm check:format`, `pnpm check:lint`, `pnpm check:types`, `pnpm check:knip` (only pre-existing hints in unrelated packages), `pnpm --filter '@portabletext/editor' test:unit` (93 files, 1415 tests) and `test:browser:chromium` (195 files, 2006 tests) all pass.

### DIFF
--- a/.changeset/shadow-root-node-type-detection.md
+++ b/.changeset/shadow-root-node-type-detection.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: identify shadow roots by node type instead of a `host` property
+
+The shadow DOM-aware DOM traversals treated any parent node carrying a `host` property as a shadow root. A `<form>` exposes its own controls as named properties, so `'host' in form` is true for any form containing an element with `id="host"` or `name="host"` — the form was mistaken for a shadow root and the upward walk stepped from the form to that control and back forever, hanging the tab.
+
+This surfaced in editors rendered inside a form that has a field named `host`: moving the selection between two editors locked the browser, because rewriting the DOM selection runs one of these traversals from a layout effect. Shadow roots are document fragments, so the node type now separates them from a form.
--- a/packages/editor/src/engine/dom/utils/dom.ts
+++ b/packages/editor/src/engine/dom/utils/dom.ts
@@ -311,8 +311,8 @@ export const closestShadowAware = (
 
     if (current.parentElement) {
       current = current.parentElement
-    } else if (current.parentNode && 'host' in current.parentNode) {
-      current = (current.parentNode as ShadowRoot).host as DOMElement
+    } else if (isShadowRoot(current.parentNode)) {
+      current = current.parentNode.host as DOMElement
     } else {
       return null
     }
@@ -345,8 +345,8 @@ export const containsShadowAware = (
     }
 
     if (current.parentNode) {
-      if ('host' in current.parentNode) {
-        current = (current.parentNode as ShadowRoot).host
+      if (isShadowRoot(current.parentNode)) {
+        current = current.parentNode.host
       } else {
         current = current.parentNode
       }
@@ -356,4 +356,23 @@ export const containsShadowAware = (
   }
 
   return false
+}
+
+/**
+ * Check if a DOM node is a shadow root.
+ *
+ * A `host` property alone does not identify one: `HTMLFormElement` exposes its own controls as named
+ * properties, so `'host' in form` is `true` for any form containing an element with `id="host"` or
+ * `name="host"`. Shadow roots are document fragments, so the node type separates the two.
+ *
+ * The node type is checked instead of `instanceof ShadowRoot` to keep working across realms and where
+ * the global is undefined.
+ */
+
+const isShadowRoot = (value: any): value is ShadowRoot => {
+  return (
+    isDOMNode(value) &&
+    value.nodeType === Node.DOCUMENT_FRAGMENT_NODE &&
+    'host' in value
+  )
 }
--- a/packages/editor/tests/dom-shadow-aware.test.ts
+++ b/packages/editor/tests/dom-shadow-aware.test.ts
@@ -1,0 +1,90 @@
+import {afterEach, describe, expect, test} from 'vitest'
+import {
+  closestShadowAware,
+  containsShadowAware,
+} from '../src/engine/dom/utils/dom'
+
+afterEach(() => {
+  document.body.innerHTML = ''
+})
+
+describe(containsShadowAware.name, () => {
+  test('finds a descendant', () => {
+    document.body.innerHTML = '<div id="foo"><span id="bar"></span></div>'
+
+    expect(
+      containsShadowAware(
+        document.getElementById('foo'),
+        document.getElementById('bar'),
+      ),
+    ).toBe(true)
+  })
+
+  test('rejects an unrelated node', () => {
+    document.body.innerHTML = '<div id="foo"></div><span id="bar"></span>'
+
+    expect(
+      containsShadowAware(
+        document.getElementById('foo'),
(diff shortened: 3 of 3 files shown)


christianhg (approved):
Thank you!

PR #3159, opened 2026-08-24
feat(plugin-sdk-value): add inline comment decorations and authoring

ryanbonial:
## Why

Sanity SDK 2.20.1 shipped comment hooks (`useComments`, `useCommentActions`), including inline text anchors on Portable Text fields. SDK apps that author content in the editor need both halves of that in the editor itself: highlights for existing comment threads, and a way to start a thread on the current selection. Today only Sanity Studio and Canvas can do this, and each hand-built the machinery privately.

This adds the machinery once, here, as two hooks. The composer UI stays with the app, the same split this package already uses for presence (`renderCursor`), and the same one Studio and Canvas chose.

## What

- `useSDKCommentDecorations({...handle, path, renderDecoration})` returns `RangeDecoration[]` for open comment threads anchored to text in this field. Anchors re-resolve when comments change; between resolutions the editor's own `onMoved` tracks edits, so there is no per-keystroke diffing. An anchor whose text is gone or rewritten beyond recognition is dropped rather than drawn on the wrong words.
- `useSDKCommentAuthoring({...handle, path})` exposes `commentableSelection` (set when the live selection can take a comment) and `createInlineComment({message})`, which captures the selection at call time and writes through `@sanity/sdk-react`.

Comments are written in the exact shape Sanity Studio stores (per selected block, the block's plain text with the selection boundaries marked by two private-use characters), so threads round-trip between an SDK app and the Studio. That format is deliberately internal to this package: the platform's comments storage is migrating, and keeping the anchor format out of the public API is what lets it change without a breaking release. The public surface is two hooks and the four types their signatures reference, mirroring the presence exports one for one.

## What to review

- `src/comments-anchoring.ts` is a direct port of Studio's `buildRangeDecorationSelectionsFromComments`, oddities preserved and annotated (the asymmetric similarity threshold, the child-separator offset walk). Its fixture suite is Studio's own, ported values and expectations.
- `src/comments-selection.ts` is the write side, new code: the editor's `getSelectedTextBlocks` already walks the selection, so this is a fraction of Studio's equivalent. Cross-container selections refuse to build rather than guessing an anchor.
- `src/plugin.sdk-comments.tsx` holds the hooks. The moved-highlight state is remembered against the comment list it was reported for, so a re-resolution wins over stale positions without an effect (the earlier draft used one; lint rightly objected).

## Testing

115 unit tests and 61 browser tests pass. The test that matters most is the round trip: what `buildStoredSelection` writes, `resolveCommentSelections` finds again at the same offsets, which is the interop contract with Studio expressed as a test. Not covered here: an end-to-end against a live dataset; the SDK side of that lives in sanity-io/sdk (see https://github.com/sanity-io/sdk/pull/1085 and https://github.com/sanity-io/sdk/pull/1087).

---

> [!NOTE]
> **Medium Risk**
> Comment anchoring is non-trivial and must stay aligned with Studio storage, but behavior is heavily unit-tested and scoped to optional hooks rather than core sync/presence paths.
> 
> **Overview**
> Adds **inline comment** support to `@portabletext/plugin-sdk-value` so SDK apps can show Studio-compatible text anchors and start new threads from the editor.
> 
> **`useSDKCommentDecorations`** loads open, anchored threads for a field via `@sanity/sdk-react`, maps stored anchors to `RangeDecoration[]` (custom `renderDecoration`), re-resolves on comment/value changes, and uses `onMoved` while typing; unrecognizable or resolved threads render no highlight.
> 
> **`useSDKCommentAuthoring`** exposes `commentableSelection` and **`createInlineComment`**, which snapshots the selection into Sanity’s stored text-anchor format and calls `createComment`.
> 
> New pure modules **`comments-anchoring`** (diff-match-patch re-anchoring, ported from Studio) and **`comments-selection`** (write path) back the hooks. Public exports include four related types. Peer dependency bumps to **`@sanity/sdk-react` ^2.20.1**; **`@sanity/diff-match-patch`** becomes a runtime dependency. README and changeset document usage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5e92fc063f87959abc7bcbdd26e7310eadad40c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/sdk-inline-comments.md
+++ b/.changeset/sdk-inline-comments.md
@@ -1,0 +1,42 @@
+---
+'@portabletext/plugin-sdk-value': minor
+---
+
+feat: add inline comment decorations and authoring
+
+Two new hooks bring Sanity's inline comments to SDK apps that embed the editor. `useSDKCommentDecorations` returns `RangeDecoration[]` for the open comment threads anchored to text in the field; pass it to `SDKPortableTextEditable` (or a raw `PortableTextEditable`) through `rangeDecorations`:
+
+```tsx
+const decorations = useSDKCommentDecorations({
+  ...documentHandle,
+  path: 'content',
+  renderDecoration: (comment) => (props) => (
+    <span data-comment-id={comment.id} style={{background: '#fef3c7'}}>
+      {props.children}
+    </span>
+  ),
+})
+
+return (
+  <SDKPortableTextEditable
+    {...documentHandle}
+    path="content"
+    rangeDecorations={decorations}
+  />
+)
+```
+
+`useSDKCommentAuthoring` is the write side: `commentableSelection` is set when the current selection can take a comment, and `createInlineComment` starts a thread anchored to that selection:
+
+```tsx
+const {commentableSelection, createInlineComment} = useSDKCommentAuthoring({
+  ...documentHandle,
+  path: 'content',
+})
+// show the composer while `commentableSelection` is set, then:
+await createInlineComment({message})
+```
+
+The types their signatures reference (`UseSDKCommentDecorationsOptions`, `RenderCommentDecorationFunction`, `UseSDKCommentAuthoringOptions`, `SDKCommentAuthoring`) are exported alongside them.
+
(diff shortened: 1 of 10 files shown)


christianhg:
Hi @ryanbonial . I'm currently working on an API to add better support for Range Decorations in plugins. I will probably land this week or next week. I wonder if this PR can wait until then? Let me know the urgency here.

ryanbonial:
@christianhg this PR is high urgency, so a merge would be appreciated. I can help with the SDK plugin rewrite when the additional support is added.

christianhg (approved):

PR #3179, opened 2026-08-26
fix: rename colliding keys before backspace merges a block into its predecessor

christianhg:
Stacked on #3154 (it supplies the transaction interpreter this change extends).

Merging a block into its predecessor with backspace re-keys any child span or annotation whose `_key` collides with one in the destination block. Today those fresh keys are minted silently before insertion, so the wire reads "block deleted, unrelated spans created": a receiving editor can connect nothing, and a collaborator whose caret sat in the merged-away block loses it (reproduced with two editors: the published selection ends up pointing into the removed block).

The fix makes the merge say what it did. Before deleting the merging block, the boundary-merge behavior renames the colliding keys in place, `set` patches on keyed `_key` paths, rewriting `marks` references when a markDef is renamed, and inserts the children under their final keys. Rename-first matters: renaming after insertion would put twin keys among siblings. Non-colliding merges emit byte-identical patches to before.

The interpreter follows the new shape by identity, never by key value: renamed nodes are tracked by their recorded path, a rename shadowing a pre-existing twin key pairs nothing, and `rekey` steps are path-scoped so a destination-block caret sitting on a twin key stays put. Adversarial probes pin each of those refusals.

One fix rides along in its own commit: undoing any `_key` rename silently dropped the rename, because the auto-generated inverse addressed the node by a path that stopped resolving once the rename applied. Inverses now target the post-rename key, pinned by a merge-undo-redo-undo round-trip.

### DIFF
--- a/.changeset/fix-key-rename-undo-inverse.md
+++ b/.changeset/fix-key-rename-undo-inverse.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: locate a renamed node by its new key when undoing a `_key` change
+
+Undoing an edit that renamed a node's own `_key` (a collision-avoiding rename ahead of a block merge, for example) left the node stuck under its new key instead of restoring the old one: the undo step tried to find the node by the key it had before the rename, which no longer resolved to anything once the rename had applied. Undoing now locates the node by its current key first, so a rename reverts cleanly along with everything else in the same undo step.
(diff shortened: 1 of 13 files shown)


christianhg:
Superseded by the un-stacked version: the emission change and both engine fixes now target main directly in #3182; the interpreter extension and the caret-follow test follow separately once #3154 merges (the branch here remains their source).

PR #3212, opened 2026-09-01
fix: support `@sanity/sdk-react@3` with normalized document handles

rexxars:
`@portabletext/plugin-sdk-value` currently rejects `@sanity/sdk-react` 3 at install time. Compiling against v3 also exposes two removed deprecated APIs: `useSanityInstance` no longer accepts a config argument, and `DocumentHandle` no longer includes the `source` alias.

This widens the peer range, reads the ambient instance without arguments, and retains `source` in the plugin's public props while translating it to `resource` before value and presence hooks receive the handle. An explicit `resource` keeps precedence when both fields are present, matching the v2 normalization behavior. The development dependency moves to v3 so the package is built and tested against the new major.

pnpm reports an upstream peer warning because `@sanity/sdk-react@3.0.0` includes `@sanity/workbench@0.1.0-alpha.24`, whose peer range still names SDK 2. The SDK React release itself publishes that dependency combination - we're working on resolving that.

### DIFF
--- a/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
+++ b/packages/plugin-sdk-value/src/plugin.sdk-value.tsx
@@ -27,12 +27,14 @@ import {
   useEditDocument,
   useSanityInstance,
   type DocumentHandle,
+  type DocumentResource,
   type EditDocumentAction,
 } from '@sanity/sdk-react'
 import {useActorRef} from '@xstate/react'
 import {useCallback} from 'react'
 import {fromCallback, setup, type AnyEventObject} from 'xstate'
 import {debug} from './debug'
+import {normalizeDocumentHandle} from './sdk-document-handle'
 
 type InsertPatch = Required<Pick<SanityPatchOperations, 'insert'>>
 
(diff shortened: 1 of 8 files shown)


packages/plugin-sdk-value/src/plugin.sdk-value.tsx
```diff
@@ -1084,6 +1086,7 @@ const valueSyncMachine = setup({
 })
 
 interface SDKValuePluginProps extends DocumentHandle {
+  source?: DocumentResource
```
christianhg, line 1089: Should this be marked as deprecated as well?

christianhg (approved):
Looks good to me.

Though not caused by this PR, I thought I would mention that the public interface for this plugin is a little messy as shown by the many `DocumentHandle`/`source` entry points. But we can probably clean that up later once Range Decorations get a proper plugin API.

packages/plugin-sdk-value/src/plugin.sdk-value.tsx
rexxars: Can do a follow-up

PR #3231, opened 2026-09-07
fix: remove the native `insertText` fast path

christianhg:
Typing a character that a behavior swallows (`on: 'insert.text'` with empty `actions`) left the character visible in the editor: the model correctly stayed unchanged, but the DOM kept it. The cause is the native insertion fast path: for a single a-z/space character at a collapsed, nonzero-offset selection, `beforeinput` skipped `preventDefault`, the browser mutated the DOM directly, and the `insert.text` behavior event ran deferred on the following `input` event. When that deferred event changed nothing, no rerender happened, so the reconciling layout effect that would repair the DOM never ran. Upstream Slate has the same bug (their open fix compensates by deleting the stray character afterwards).

This removes the fast path instead: `insertText` now always prevents the default and sends `insert.text` synchronously, the route every other input already took (punctuation, digits, offset 0, IME, Android). The fast-path guards, the deferred-operation machinery, and a dead `DOMText` export go with it; the diff is net negative. Pinned by a browser test asserting a swallowed character reaches neither the model nor the DOM, red on the fast-path code.

The semantic delta worth scrutiny: previously-native keystrokes now render through the model. Typing latency measured at parity (40 keystrokes into a 300-block document, headless chromium, 3 runs: 193/236/204 ms with the fast path, 138/188/198 ms without). A deleted comment claimed the deferral kept autocorrect and spellcheck working; the full chromium (2042 tests) and firefox (2022) suites pass, but spellcheck interaction has no pinning test, so that path is verified only by suite coverage. WebKit did not run: the browser session fails to connect on this machine, reproduced on an untouched `main` checkout.

---

> [!NOTE]
> **Medium Risk**
> Changes core text-input handling for all single-character typing; previously-native keystrokes now render only through the model, with spellcheck/autocorrect behavior verified only by existing suite coverage rather than a dedicated test.
> 
> **Overview**
> Removes the **native `insertText` fast path** so single-character typing no longer bypasses the editor model. **`beforeinput` for `insertText` always calls `preventDefault()`**, and **`insert.text` behavior events are sent synchronously** instead of being deferred until the next `input` event.
> 
> That fixes a bug where a Behavior that **swallows** `insert.text` (empty `actions`) left a **ghost character in the DOM** while the document model stayed correct—the browser had already inserted the character natively, and no rerender ran to reconcile.
> 
> Also **drops the deferred-operation queue**, the Chrome-specific guards (anchors, `pre` + tabs, a–z/space-only), and the unused **`DOMText` export**. A **browser test** asserts swallowed `x` never appears in the model or `textContent`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bda594d059fc649a667a7af3449c3f98b4bc6a2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/remove-native-insert-fast-path.md
+++ b/.changeset/remove-native-insert-fast-path.md
@@ -1,0 +1,7 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: remove the native `insertText` fast path
+
+Typing a character that a Behavior swallows no longer leaves the character visible in the editor: all text insertion now flows through the model, so a swallowed `insert.text` event never leaves a stray character stuck in the DOM.
--- a/packages/editor/src/engine/dom/utils/dom.ts
+++ b/packages/editor/src/engine/dom/utils/dom.ts
@@ -11,11 +11,6 @@ import {DOMEditor} from '../plugin/dom-editor'
 type DOMNode = globalThis.Node
 type DOMComment = globalThis.Comment
 type DOMElement = globalThis.Element
-// DOMText is used as a value (instanceof) in dom-editor.ts, so we need both
-// the type alias and a const reference to the global constructor.
-// oxlint-disable-next-line no-redeclare -- type and value live in separate TypeScript namespaces
-type DOMText = globalThis.Text
-const DOMText = globalThis.Text
 type DOMRange = globalThis.Range
 type DOMSelection = globalThis.Selection
 type DOMStaticRange = globalThis.StaticRange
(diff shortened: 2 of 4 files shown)


christianhg:
Closing in favor of the decided input-pipeline design: native plain-text insertion stays (no surviving editor blocks plain typing, and `preventDefault` is not universally honored, so a repair path must exist regardless), and this bug class gets closed by a pipeline repair primitive that reverts the browser-painted range when the resolved actions do not execute the default insert. The pinning test here transfers to that work unchanged; the branch stays for the test and the typing-latency measurement.

PR #3247, opened 2026-09-08
fix: delete auto-resolve validation repairs in favor of engine normalization

christianhg:
With #3246, the editor emits its structural repair patches the moment a value settles. That left `validateValue`'s `autoResolve` resolutions as a second repairer running ahead of the engine: four rules (missing block `_key`, missing or empty `children`, orphaned `markDefs`, missing child `_key`) repaired blocks before ingestion and reported through the sync actor's out-of-band `patch` channel, with their own patch shapes and their own addressing quirks.

This PR deletes them and makes engine normalization the sole repairer. Validation passes mechanically fixable shapes through untouched; the engine repairs them on intake and its patches emit in engine shapes: a keyless block gets a minimal `set` on its minted `_key` (previously a whole-block `set`), an empty text block gets its placeholder span as an `insert` before `children[0]`. The sync actor's `patch` channel and its reporting machinery go with the rules. The invalid-value flow is untouched for defects that genuinely need a human, and its remaining resolutions now resolve their block anchors through `nodeSegment`, so a keyless block with a non-mechanical defect is addressed by its numeric index instead of a `{_key: undefined}` segment that keyed matching would resolve to the first keyless sibling (pinned red-on-old).

Raw input reaches `updateBlock` for the first time, so it gains a wholesale-set fallback: children arrays with keyless members, or a raw empty `children` replacing populated children, are set wholesale and normalization mints the keys; no `{_key: undefined}` segment is ever applied (pinned by unit tests asserting the full applied-operation streams).

Two deliberate deltas. Orphaned `markDefs` are no longer pruned at intake: the engine's prune is local-only by design, so orphans go when a local edit next dirties the block, emitted as a `set` of the filtered array (a read-only editor never prunes them). And sync `insert` operations now carry the raw block, with normalization repairs following as their own operations. `InvalidValueResolution.autoResolve` is deprecated rather than removed, never set.

One caveat: the empty-block placeholder span now travels as `insert ... before children[0]` where the deleted rule used `insert ... after children[0]`; both anchor index 0 of an empty array, and whether Content Lake treats the two anchors identically on an empty array is not verifiable from this repo. Client-side application (`applyAll`) is pinned equivalent.

Stacked on #3246 and merges after it; the base branch is `emit-repair-patches-immediately`.

### DIFF
--- a/.changeset/delete-auto-resolve-repairs.md
+++ b/.changeset/delete-auto-resolve-repairs.md
@@ -1,0 +1,9 @@
+---
+'@portabletext/editor': patch
+---
+
+fix: delete auto-resolve validation repairs in favor of engine normalization
+
+The editor's own normalization now repairs every mechanically fixable defect in a synced value (missing or empty `children`, missing block and child `_key`s), and the repair patches take the engine's shapes: a keyless block gets a minimal `set` on its `_key` field instead of a whole-block `set`, and an empty text block gets its placeholder span as an `insert` before `children[0]`.
+
+Two deltas ride along. Orphaned `markDefs` are no longer pruned when a value enters the editor; they are pruned when a local edit next touches the block, emitted as a `set` of the filtered `markDefs` array. And `InvalidValueResolution.autoResolve` is deprecated and never set: defects the editor can fix itself no longer produce a resolution, so the invalid-value flow only fires for defects that genuinely need a human.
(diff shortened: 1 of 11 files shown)


christianhg:
Folded into #3246 as its third commit; one PR, same commits.

PR #3262, opened 2026-09-10
chore: publish pkg.pr.new preview packages on labeled pull requests

pepahlavacek:
## Description

Add a "Publish Preview Packages" workflow that runs on pull requests carrying the `trigger: preview` label, builds `packages/*`, and hands them to `pkg-pr-new publish`. The CLI is a root devDependency executed from the lockfile rather than fetched with `pnpm dlx`, so the version that runs in CI is pinned. `--pnpm` is required because the workspace uses `catalog:` specifiers, which `npm pack` cannot resolve.

The CLI authenticates through the pkg.pr.new GitHub App using the workflow run metadata; the app posts the PR comment, so the job needs no `GITHUB_TOKEN` and only `contents: read`.

Peer dependencies are not rewritten to preview URLs, so a plugin preview still resolves `@portabletext/editor` from npm unless the editor preview is installed alongside it. The README documents this.

### DIFF
--- a/.github/workflows/pkg-pr-new.yml
+++ b/.github/workflows/pkg-pr-new.yml
@@ -1,0 +1,79 @@
+name: Publish Preview Packages
+
+on:
+  pull_request:
+    types: [labeled, synchronize, opened, reopened]
+
+concurrency:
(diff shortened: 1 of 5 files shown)


christianhg (approved):
Nice

PR #3271, opened 2026-09-14
fix: restore `data-slate-editor` on the editable element so Grammarly recognizes it

christianhg:
Accepting a Grammarly correction persists again. Grammarly fingerprints editors by DOM attributes: with `data-slate-editor` present it applies corrections through synthetic `beforeinput` events, which the editor handles. Since v8.0.0 removed the Slate-era `data-slate-*` aliases, Grammarly fell back to a DOM-mutation channel the editor cannot observe, so corrections looked applied but never reached the document and reverted.

The attribute is a temporary workaround, not public API: it will be removed again, without warning, once the editor handles that fallback channel without it. Do not style or query against it.

Only this one alias comes back; the rest stay removed. The bare JSX boolean renders as `data-slate-editor="true"`, which is the exact value Grammarly's fingerprint checks (verified against `react-dom/server` output). Other extensions that use the same `document.execCommand` fallback are not helped by this; handling that channel is the durable fix and is tracked separately. The DOM structure suite passes unchanged and pins nothing about the attribute's absence.

---

> [!NOTE]
> **Low Risk**
> Single boolean DOM attribute on the editable root with inline documentation; no changes to input handling logic beyond re-enabling Grammarly’s preferred integration path.
> 
> **Overview**
> Restores the **`data-slate-editor`** DOM attribute on the `<Editable>` root (alongside existing `data-pt-editor`) so Grammarly again treats the field as a Slate-style editor and applies suggestions via synthetic **`beforeinput`** events the engine already handles.
> 
> After v8 dropped Slate-era `data-slate-*` aliases, Grammarly used a **`document.execCommand`** path with no `beforeinput`, so accepted corrections looked applied in the DOM but did not update the document and reverted. The change is documented as a **temporary, non-API** workaround (patch changeset); it may be removed without notice once the editor supports that fallback channel. Integrators should not style or query on this attribute.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a63a479a0d6e8eeb7641c45a00e80b36521a120f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


bjoerge (approved):
Thank you for this!

PR #3275, opened 2026-09-15
test: pin store convergence for the emitted mutation stream

christianhg:
A host that persists the editor's emitted mutations (a form store, a patch-applying backend) relies on an invariant no test asserted directly: replaying the emitted stream converges to the editor's own value. When it breaks, the wire shows a whole-field `unset` followed by patches targeting `_key`s the unset destroyed, and a store that treats zero-match patches as silent no-ops loses the content without an error.

The rig is a reference store that applies each emitted patch with `applyAll` and drops any patch it cannot apply, matching those store semantics, plus a mirror host feeding the store's value back as `update value`. One scenario pins the healthy clear-and-retype path green. Two `test.fails` scenarios document live divergences: a stale echo of the pre-clear value syncing back in after the editor's own `unset([])` flushed, after which edits emit bare `diffMatchPatch`es into the destroyed field while the store stays `undefined`; and a clear held by the typing debounce through a read-only window, where a restore arriving through `update value` during the window never reaches the editor and the stale `unset([])` flushes after the editor becomes editable, destroying the restored value. A fix for either flips its scenario to a plain `test`.


christianhg:
Folded into #3276 as a single PR (rig + fix).

